PR #2, opened 2020-05-06
ComputeStatsGeotrellisAdapter improvements

soxofaan:
internal ticket EP-3399

- reduce boilerplate and code duplication
- add "vector file" support for "histogram" and "sd"
- harmonize overloaded compute functions for "mean", "median", "histogram" and "sd"

### DIFF
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/ComputeStatsGeotrellisAdapter.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/ComputeStatsGeotrellisAdapter.scala
@@ -111,42 +111,6 @@ class ComputeStatsGeotrellisAdapter(zookeepers: String, accumuloInstanceName: St
       statisticsWriter.close()
   }
 
-  def compute_histogram_time_series(product_id: String, polygon_wkt: String, polygon_srs: String,
-                                    from_date: String, to_date: String, zoom: Int, band_index: Int):
-  JMap[String, JMap[Double, Long]] = { // date -> value/count
-    val computeStatsGeotrellis = new ComputeStatsGeotrellis(layersConfig(band_index))
-
-    val polygon = ProjectedPolygons.parsePolygonWkt(polygon_wkt)
-    val crs: CRS = CRS.fromName(polygon_srs)
-    val startDate: ZonedDateTime = ZonedDateTime.parse(from_date)
-    val endDate: ZonedDateTime = ZonedDateTime.parse(to_date)
-
-    val histograms = computeStatsGeotrellis.computeHistogramTimeSeries(product_id, polygon, crs, startDate, endDate, zoom, sc)
-
-    histograms
-      .map { case (temporalKey, histogram) => (isoFormat(temporalKey.time), toMap(histogram)) }
-      .collectAsMap()
-      .asJava
-  }
-
-  def compute_histogram_time_series_from_datacube(datacube:MultibandTileLayerRDD[SpaceTimeKey], polygon_wkt: String, polygon_srs: String,
-                                    from_date: String, to_date: String, band_index: Int):
-  JMap[String, JList[JMap[Double, Long]]] = { // date -> band -> value/count
-    val computeStatsGeotrellis = new ComputeStatsGeotrellis(layersConfig(band_index))
-
-    val polygon = ProjectedPolygons.parsePolygonWkt(polygon_wkt)
-    val crs: CRS = CRS.fromName(polygon_srs)
-    val startDate: ZonedDateTime = ZonedDateTime.parse(from_date)
-    val endDate: ZonedDateTime = ZonedDateTime.parse(to_date)
-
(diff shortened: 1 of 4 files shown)


soxofaan:
Now also pushed commit with new/updated unit tests

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/ComputeStatsGeotrellisAdapter.scala
```diff
@@ -29,6 +29,108 @@ import scala.collection.JavaConverters._
...
+            .flatMap(asMultiPolygons)
+            .toArray
+      }
+    } finally src.close()
+
+    ProjectedPolygons(multiPolygons, LatLng)
+  }
+
+}
+
```
bossie, line 133: In Scala, unlike Java, you can have multiple public classes within a single file but `ProjectedPolygons` is large enough to warrant a separate file; put `ProjectedPolygons` and its companion object into a file called ProjectedPolygons.scala next to this file.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/ComputeStatsGeotrellisAdapter.scala
```diff
@@ -150,96 +236,17 @@ class ComputeStatsGeotrellisAdapter(zookeepers: String, accumuloInstanceName: St
...
-          featureCollection.getAllGeometries()
-            .flatMap(asMultiPolygons)
-            .toArray
-      }
-    } finally src.close()
-
-    (multiPolygons, LatLng)
-  }
-
   def compute_histogram_time_series(product_id: String, polygon_wkt: String, polygon_srs: String,
```
bossie, line 244: Since `ProjectedPolygons` becomes the primary input abstraction of these methods, we might want to reduce the surface area of the interface and get rid of methods that accept `polygon_wkt(s)` and `polygon(s)_srs`; unit tests can be adapted by calling the ProjectedPolygons primary constructor.

`compute_histogram_time_series` and `compute_histogram_time_series_from_datacube` look like they can be removed, eliminating the need to expose `ProjectedPolygons#parsePolygonWkt` as public: you can make it private instead.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/ComputeStatsGeotrellisAdapter.scala
```diff
@@ -150,96 +236,17 @@ class ComputeStatsGeotrellisAdapter(zookeepers: String, accumuloInstanceName: St
...
                                     from_date: String, to_date: String, zoom: Int, band_index: Int):
   JMap[String, JMap[Double, Long]] = { // date -> value/count
     val computeStatsGeotrellis = new ComputeStatsGeotrellis(layersConfig(band_index))
 
-    val polygon = parsePolygonWkt(polygon_wkt)
+    val polygon = ProjectedPolygons.parsePolygonWkt(polygon_wkt)
```
bossie, line 249: `compute_histogram_time_series` and `compute_histogram_time_series_from_datacube` look like they can be removed, eliminating the need to expose `ProjectedPolygons#parsePolygonWkt` as public: you can make it private instead.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/ComputeStatsGeotrellisAdapter.scala
soxofaan: indeed, `ProjectedPolygons` turned out larger than I anticipated, 
will move to new file

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/ComputeStatsGeotrellisAdapter.scala
soxofaan: indeed, I was thinking the same, but was not sure if I could just drop existing methods

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/ComputeStatsGeotrellisAdapter.scala
soxofaan: ok will do

soxofaan:
pushed a couple of commits based on review of @bossie

soxofaan:
rebased (minor cleanup) on top of latest develop and force pushed PR branch

jdries:
All looks like good improvements. Have we verified that this will also work from the Python side? Ideally, we can keep the Python-Java interface backwards compatible.
Also good to consider that Py4J mostly supports plain java, but can have issues with scala constructs. I did not immediately see something that would clearly be an issue.

soxofaan:
the python side is indeed covered by https://github.com/Open-EO/openeo-geopyspark-driver/pull/40

PR #4, opened 2020-05-25
Add support for v1-style processes `or`, `and`, ..., `sum`, `add`, `divide`, ...

soxofaan:
v0.4 defines most processes like `or`, `and`, `sum`, `divide` as "reduce" style processes, which is bit weird (especially for subtract and divide)

v1.0 improves that by defining logical processes, `subtract`, `divide` as "xy" processes, and adds `add` `multiply` as "xy" processes.

This PR makes sure we support both approaches properly (based on the provided arguments: `expressions` or `data` or `x`/`y`?)

largest part of diff of this PR are unit tests (more coverage, more code reuse, deeper asserts)


jdries:
Approved, great progress on unit test quality!

PR #5, opened 2020-10-06
Temporary fix for retries logic

sinergise-anze:
The Sentinel Hub rate limiting provides `Retry-After` response header which advises clients on how long they should wait for before issuing a next request, to avoid being rate-limited. However, the assumption here is that there is only a _single_ consumer. When there are multiple consumers (as in this case) only one of them will be able to get the response while others will be (again) rejected.

Additionally, the code should react differently depending on the response status code:
- `429`: retry later
- other `4xx`: abort, there is some problem with the request
- `5xx`: retry later; backoff times should be larger to avoid compounding to service troubles

This PR is _not_ yet a proper solution - we should discuss the architecture of the system to find a better one. However, it should increase likelihood that each worker will be able to finish their work, because as soon as errors start, some workers will back off for non-trivial amount of time, leaving others enough space to continue, and hopefully not timing out themselves yet. 

Please see this PR as a suggestion for a temporary fix (feel free to disregard and close too). We can discuss better alternatives live.


jdries:
Ok, this already seems to contain some initial improvements, so I'll merge it. That will allow us to test it again.

PR #6, opened 2020-10-06
Replace WCS with Process API (Sentinel Hub)

sinergise-anze:
This PR replaces WCS requests with [Sentinel Hub Process API](https://docs.sentinel-hub.com/api/latest/api/process/) ([API reference](https://docs.sentinel-hub.com/api/latest/reference/#tag/process)). This API is much more powerful and even allows things such as data fusion between datasets.

There are a couple of things that this PR _doesn't_ solve:

- `authToken` (see [Sentinel Hub authentication](https://docs.sentinel-hub.com/api/latest/api/overview/authentication/)) must be somehow obtained from the service and distributed to workers. Instead of assuming, I have simply created an `authToken` variable, but this was of course only useful for very basic testing.

- Sentinel Hub API (both OGC WMS/WCS and Process API) allows fetching multiple bands in a single request, which is much more efficient. As this is out of scope for fetching data (it needs to be consumed properly too) I have simply indicated where the bands should be specified if you wish to fetch multiple at once (see use of `band` and `nBands` variables withing the `evalscript` definition).

- the original code fetched TIFF 32f, which is not very efficient, especially as the original data doesn't have such resolution either. I would suggest changing this to TIFF 16. **IMPORTANT:** if you do this, make sure to also change the `evalscript` so that it returns data between min / max possible value, otherwise the images will only have values `0` and `1`.

- I was able to run this code by commenting out `@Ignore` on one of the tests (`testSentinel2L2A`); however, while it seemed to work, there could be some bugs. The test also didn't fail when the requests failed so I am not sure it is finished. Please test additionally.

### DIFF
--- a/geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
+++ b/geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
@@ -17,11 +17,11 @@ import org.openeo.geotrellissentinelhub.bands._
 
 import scala.collection.JavaConverters._
 
-abstract class PyramidFactory[B <: Band](val uuid: String) extends Serializable {
+abstract class PyramidFactory[B <: Band]() extends Serializable {
   private val maxZoom = 14
-  protected val endpoint = "https://services.sentinel-hub.com"
   val allBands: Seq[B] = Seq()
-  
+  val datasetId: String = ""
+
   private def sequentialDates(from: ZonedDateTime): Stream[ZonedDateTime] = from #:: sequentialDates(from plusDays 1)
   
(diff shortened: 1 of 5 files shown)


jdries:
Nice, we definitely want to integrate this to get optimal performance.
First step will be to find a way to properly create tokens, and handle expiration. Does SentinelHub already support certain external IDP's, or is it purely shub accounts?

sinergise-anze:
Just SH accounts, as far as I know. Do you have something specific in mind?

I have included a [link](https://docs.sentinel-hub.com/api/latest/api/overview/authentication/) to authentication docs - should be pretty straightforward. It might be easiest for you if each worker requested the tokens by itself and refreshed them when / before they expire (you can have multiple auth tokens active at once, so this is not a problem). It might be wise to use some "warm up" period though, espeially if you want to have a lot of workers.

jdries:
In the platform we'll need to support eosc users, through the EGI check-in IDP:
https://www.egi.eu/services/check-in/
But let's get it to work with a simple test account first, and then discuss on best options for dealing with user management.

sinergise-anze:
Sounds like a good plan, yes.

Let me know if there is anything else I can do to help.

PR #17, opened 2020-12-02
Feature/ep 3627

jdries:


jdries:
The 'layer' method still seems to have a bit of overlap with readMultibandTileLayer in FileLayerProvider. Would it be possible to further align that code?

PR #18, opened 2020-12-08
Feature/creo open search

laxiwuka:


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/file/FileRDDFactory.scala
```diff
@@ -27,7 +26,7 @@ import scala.collection.JavaConverters._
 class FileRDDFactory(openSearchCollectionId: String, openSearchLinkTitles: util.List[String],attributeValues: util.Map[String, Any] = util.Collections.emptyMap(),correlationId: String = "") {
 
   private val maxSpatialResolution = CellSize(10, 10)
-  protected val openSearch: OpenSearch = OpenSearch(new URL("http://oscars-01.vgt.vito.be:8080"))
+  protected val openSearch: OscarsOpenSearch = OscarsOpenSearch(new URL("http://oscars-01.vgt.vito.be:8080"))
```
jdries, line 29: The type here should remain OpenSearch, not a specific subclass, so we can ensure that we working against the abstract base class.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/OpenSearch.scala
```diff
@@ -21,75 +21,25 @@ import scala.collection.Map
 object OpenSearch {
   private val logger = LoggerFactory.getLogger(classOf[OpenSearch])
   private val requestCounter = new AtomicLong
-
-  def apply(endpoint: URL) = new OpenSearch(endpoint)
```
jdries, line 25: This could also have been a factory method that tries to return the 'right' type of opensearch?

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/OscarsOpenSearch.scala
```diff
@@ -0,0 +1,69 @@
...
+import scala.collection.Map
+
+object OscarsOpenSearch {
+  def apply(endpoint: URL) = new OscarsOpenSearch(endpoint)
+}
+
+class OscarsOpenSearch(endpoint: URL) extends OpenSearch {
+  def getProducts(collectionId: String, from: LocalDate, to: LocalDate, bbox: ProjectedExtent,
+                            correlationId: String = "", attributeValues: Map[String, Any] = Map()): Seq[Feature] = {
+    val endOfDay = OffsetTime.of(23, 59, 59, 999999999, UTC)
```
jdries, line 21: Isn't this implementation the same for Creo? (I could be missing a detail here.)

jdries (changes requested):

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/OscarsOpenSearch.scala
laxiwuka: In creo we also have to pass the processingLevel somehow. Not sure what the best approach is to get rid of this.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/file/FileRDDFactory.scala
laxiwuka: Only problem to change this is that the method signatures differ because we need to pass processingLevel in creo. Maybe that can be a constructor parameter?

The getCollections() return type is also not the same because the output is different in oscars and creo. Maybe we don't need that function in creo, because the one in oscars is only used to fetch the extent for the metadata and that information isn't available in the creo collections response.

jdries: One solution: pass parameters in a more generic way, like a map.
Other option: also add processinglevel to oscars implementation and not use it?

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/OpenSearchResponses.scala
```diff
@@ -41,4 +41,48 @@ object OpenSearchResponses {
...
+            resolution = c.downField("properties").downField("resolution").as[Int].toOption
+          } yield {
+            val extent = geometry.toString().parseGeoJson[Geometry].extent
+
+            Feature(id, extent, nominalDate, links, resolution)
+          }
+        }
+      }
+
+      implicit val decodeFeatureCollecion: Decoder[FeatureCollection] = new Decoder[FeatureCollection] {
```
soxofaan, line 63: minor typo in `decodeFeatureCollecion`

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/file/FileRDDFactory.scala
soxofaan: Maybe I'm misunderstanding, but FileRDDFactory should also be usable against creo open search in some way, right?

jdries: Looks better now!

jdries (approved):
Looks better now, I would merge, so we can look at how to use on DIAS.

PR #20, opened 2021-05-06
EP-3832: Move OpenSearch clients to the gwc-geotrellis project

JeroenVerstraelen:
Moved the OpenSearch clients to the gwc-geotrellis project:

- Adjusted the opensearch code in geotrellis-seeder module
- Added gwc-geotrellis-core-0.11.0-SNAPSHOT dependency to openeo-geotrellis module
- Replaced opensearch clients in openeo-geotrellis module with the ones from the gwc-geotrellis-core project

Note:
The build passes all tests and checks but fails on the PyramidFactoryTest.testLandsat8 test. This happens because https://services.sentinel-hub.com appears to be down at the moment. The current build on the develop branch appears to have the same issue.


jdries (approved):
approved, go ahead and merge!

PR #21, opened 2021-05-06
Bugfix/ep 3828 new copernicus 30 layer returns alignment error

JeroenVerstraelen:
Solved bug where DEM layer introduced an alignment error:
- Made noResampling more robust to rounding errors
- Made noResampling flag depend on the type of projection system 
- Fixed an edge case if rastersource is on the boundary of the extent


jdries (changes requested):
is it an option to check the geotiff that comes out of the unit test?
If it is not too large, we can even check it into the repo?

PR #23, opened 2021-05-21
Feature/ep 3844 space partitioner for sparse data

jdries:


openeo-geotrellis/src/test/scala/org/openeo/geotrellis/layers/FileLayerProviderTest.scala
```diff
@@ -74,4 +78,105 @@ class FileLayerProviderTest {
 
     assertEquals(25, physicalMean)
   }
+
+  @Test
+  def sparsePartitionerTest(): Unit = {
```
jdries, line 83: is it an option to split up this test into multiple smaller tests?

geotrellis-common/src/main/scala/org/openeo/geotrelliscommon/package.scala
```diff
@@ -1,10 +1,26 @@
 package org.openeo
 
-import org.openeo.geotrelliscommon.zcurve.SfCurveZSpaceTimeKeyIndex
 import geotrellis.layer.SpaceTimeKey
 import geotrellis.spark.partition.PartitionerIndex
+import org.apache.spark.rdd.RDD
```
jdries, line 5: Would it also be an option to avoid a dependency on spark? Spark is a fairly heavy thing to depend on, and we only introduce it here for one line. 
Only thing is this logic with the keyindex, that you probably want to have in here. Maybe we can define an indexing function in geotrellis common, that is then used in a module that translates the RDD of spacetimekey into a simple list of indices?

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/layers/FileLayerProviderTest.scala
JeroenVerstraelen: I used one large test with a large dataset/extent to really test the difference between the two partitioners when there are many partitions. I was reluctant to split it up because then the large RDD would have to be loaded in for every test. 

However, looking at it again, I can definitely split the test up and use multiple smaller datasets.

PR #25, opened 2021-08-03
EP-3819: add initial support for udfs with off-heap memory tiles

JeroenVerstraelen:
Initial support to execute UDFs with off-heap memory tiles:

- Added JEP library as a dependency 
- Added a Udf class with a public runUserCode method that can be accessed from Python
- Added simple Udf tests

### DIFF
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/udf/Udf.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/udf/Udf.scala
@@ -12,14 +12,12 @@ object Udf {
 
   private val defaultImports =
     """
+      |import collections
       |import numpy as np
       |import xarray as xr
       |import openeo.metadata
-      |try:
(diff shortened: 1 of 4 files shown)


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/udf/Udf.scala
```diff
@@ -0,0 +1,171 @@
...
+
+object Udf {
+
+  private val defaultImports =
+    """
+      |import numpy as np
+      |import xarray as xr
+      |import openeo.metadata
+      |try:
+      |    from openeo_udf.api.base import UdfData, SpatialExtent
```
jdries, line 19: These openeo udf classes were recently moved into the Python client, and can be found under openeo.api.udf. That avoids having to depend on openeo_udf.

jdries (changes requested):
Minor remark added, will be interesting to see this in action!

jdries:
Nice, I merged it!

PR #26, opened 2021-08-26
EP-3985: add maxSpatialResolution parameter for Sentinelhub pyramid c…

JeroenVerstraelen:
Add maxSpatialResolution parameter when creating Sentinelhub pyramid.
This PR is linked with [the one in the openeo-geopyspark-driver project](https://github.com/Open-EO/openeo-geopyspark-driver/pull/104).

### DIFF
--- a/geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
+++ b/geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
@@ -52,8 +52,8 @@ object PyramidFactory {
 
   @deprecated("include a collectionId")
   def rateLimited(endpoint: String, datasetId: String, clientId: String, clientSecret: String,
-                  processingOptions: util.Map[String, Any], sampleType: SampleType, maxSpatialResolution: CellSize = CellSize(10,10)): PyramidFactory =
-    rateLimited(endpoint, collectionId = null, datasetId, clientId, clientSecret, processingOptions, sampleType, maxSpatialResolution)
+                  processingOptions: util.Map[String, Any], sampleType: SampleType): PyramidFactory =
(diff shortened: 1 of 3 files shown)


jdries:
There's a compilation error?

JeroenVerstraelen:
My bad, those errors should now be gone.

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
```diff
@@ -47,9 +46,9 @@ object PyramidFactory {
   // convenience method for Python client
   // TODO: change name? A 429 response makes it rate-limited anyway.
   def rateLimited(endpoint: String, collectionId: String, datasetId: String, clientId: String, clientSecret: String,
-                  processingOptions: util.Map[String, Any], sampleType: SampleType): PyramidFactory =
+                  processingOptions: util.Map[String, Any], sampleType: SampleType, maxSpatialResolution: CellSize = CellSize(10,10)): PyramidFactory =
```
jdries, line 49: Is this going to be backwards compatible when called from python? I'm not sure if the default arguments work in there.

JeroenVerstraelen: The default arguments should be backwards compatible now. The build fails due to the same two tests that are broken in the develop branch: 
- org.openeo.geotrellissentinelhub.S3ServiceTest.download_stac_dataCanHandleBatchJobRetries
- org.openeo.geotrellissentinelhub.S3ServiceTest.download_stac_data

We should be able to merge without issues.

jdries: Great!

PR #27, opened 2021-08-26
EP-3967: Sentinel-2 L1C cloudmasks on CreoDIAS

JeroenVerstraelen:
* Added GDALCloudRasterSource
This source is used when a .jp2 file contains FineCloudMask_Tile1_Data as a feature
This source handles reading/handling polygons from a cloud file when the "mask_l1c" masking strategy is defined. 
This source also reads the extent from the S2_Level-1C_Tile1_Metadata file, to avoid reading from the .jp2 file on the driver side. Note: This is used to check if the cloud polygons cover the entire extent.

* Added filtering of raster sources 
When a complete raster source is fully clouded we can save a lot of IO by ignoring it.

* Added filtering of raster regions
When a raster region is fully clouded it is filtered out of the list of available regions to save IO.
This uses the dilated and merged polygons from the rastersource filtering step to avoid having to recompute this for every requested region (merging many cloud polygons is a computationally expensive task). 

* Fixed issue with CRS returning empty (and defaulting to LatLng) when read from Sentinelhub L1C .jp2 files (causing an exception during reprojection).
GDALCloudRasterSources currently get this CRS from the cloud file to avoid this issue.

TODO
Note that currently the default dilation distance is set to 250 meters by using some rough estimations. These calculations can be found in the comments of the JIRA story. This distance sounds rather large but I'm not sure if sources/regions would actually get pruned if it was lower as there seems to be a large difference between the CLOUD_PERCENTAGE field found in the metadata file and the percentage calculated by dividing the area of the cloud polygons by the area of the source extent.  

JENKINS BUILD
The build currently fails because some unrelated tests are unable to connect to their respective backend server. 
 (sobloo.eu, sentinelhub)

EXTRA MASKING STEP 6/09/2021
Extra masking step has been added. Lower dilation distances of ~5 meters seem to work fine to prune raster regions and for masking (while ~250m was required to prune entire raster sources). The merging of the cloud polygons inside the cloud mask metadata file has a large impact on performance, it happens once per entire raster source but for 'testL1CMultibandTileMask', the l1c_mask increases test time from ~32s to ~2min 30s.

JENKINS BUILD 6/09/2021
Some unrelated tests appear to fail with org.openeo.geotrelliss3.CatalogException: Status: 504, Message: Gateway Timeout


jdries:
Quite a PR! Let's merge it, so we can also test on Creo. No code updates required for now, but github indicates a conflict.

JeroenVerstraelen:
> Quite a PR! Let's merge it, so we can also test on Creo. No code updates required for now, but github indicates a conflict.

Great! The merge conflicts have been resolved and the Jenkins build has been restarted.

PR #29, opened 2021-09-23
Feature/ep 4015 support first and last processes

JeroenVerstraelen:


openeo-geotrellis/src/test/java/org/openeo/geotrellis/TestOpenEOProcessScriptBuilder.java
```diff
@@ -694,6 +694,46 @@ public void testIfWithRejectCube() {
...
+        builder.expressionEnd("last",arguments);
+
+        Function1<Seq<Tile>, Seq<Tile>> transformation = builder.generateFunction();
+        ByteArrayTile tile0 = ByteConstantNoDataArrayTile.fill((byte) 10, 4, 4);
+        ByteArrayTile tile1 = ByteConstantNoDataArrayTile.fill((byte) 5, 4, 4);
+        Seq<Tile> result = transformation.apply(JavaConversions.asScalaBuffer(Arrays.asList(tile0, tile1)));
+        Tile res = result.apply(0);
+        assertTileEquals(tile1, res);
+    }
+
```
soxofaan, line 736: Can you also add a test with nodata values in the input and check if the `ignore_nodata` option works properly?

JeroenVerstraelen: Tests have been added.

JeroenVerstraelen:
- Ensured that the first/last process do not change the cell type to Double
- Increased the performance of the first/last process
- Added extra tests for the ignore_nodata parameter

jdries:
Looks good!

Issue #102, opened 2022-12-11
aggregate_temporal or linear interpolation seems to convert numbers to floats


jdries:
Not very clear where it happens exactly, but the code below seems to convert shorts into floats, which is not strictly necessary.

```
        S1bands = S1bands.aggregate_temporal_period(period="dekad",reducer="mean")
        S1bands = S1bands.apply_dimension(dimension="t",process="array_interpolate_linear")
```

jdries:
@EmileSonneveld  For this one, we could write a more generic unit test.
On the input side, we need a spatiotemporal datacube (with nodata). This input then needs to be converted into various other data types, that we all test separately.
This is one example of potential input, based on real data:
https://github.com/Open-EO/openeo-geotrellis-extensions/blob/19df3be87e0bb4f32294c270fd1aade063270221/openeo-geotrellis/src/test/scala/org/openeo/geotrellis/LayerFixtures.scala#L133

To be discussed!

EmileSonneveld:
In this test, the result stays in in Int type. Under the hood the maps get temporary converted to doubles type, but they get casted back again.

```python
import openeo

connection = openeo.connect("https://openeo.vito.be").authenticate_oidc()

load1 = connection.load_collection(
    collection_id="SENTINEL2_L2A",
    spatial_extent={"west": 5.07, "east": 5.08, "south": 51.215, "north": 51.22},
    temporal_extent=["2022-10-01T00:00:00Z", "2022-12-01T00:00:00Z"], bands=["B02"]
)

load1 = load1.aggregate_temporal_period(period="dekad", reducer="mean")
load1 = load1.apply_dimension(dimension="t", process="array_interpolate_linear")

print("load1 made: ", load1)
print("download start")
load1.download("should_be_UInt16.nc")
print("download done")

```
![image](https://user-images.githubusercontent.com/2298426/211267688-df093dcb-cede-4d7c-97ee-91e5e28624eb.png)

EmileSonneveld:
@jdries , I also made more geotrellis-common tests run. Now all Specs in this repo should run. (tested locally with `mvn test`)
Does this ticket look ok?

jdries:
ok, closing!

PR #107, opened 2023-01-11
#102 types test

EmileSonneveld:


openeo-geotrellis/src/test/scala/org/openeo/geotrellis/OpenEOProcessesSpec.scala
```diff
@@ -291,6 +298,71 @@ class OpenEOProcessesSpec extends RasterMatchers {
...
+    val outDir = "/tmp/aggregateTemporalTest/"
+    Files.createDirectories(Paths.get(outDir))
+    List(
+      DoublePixelType(),
+      FloatPixelType(),
+      IntPixelType(),
+      ShortPixelType(),
+      BytePixelType(),
+      BitPixelType()
+    ).foreach(pixelType => {
```
jdries, line 313: a for loop like this can be turned into a parametrized test, which gives more convenient output

jdries (changes requested):
ok, reviewed, one comment!

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/OpenEOProcessesSpec.scala
EmileSonneveld: Used `@ParameterizedTest` way.

PR #111, opened 2023-01-18
96 merge cubes

EmileSonneveld:


EmileSonneveld:
I realized the mask test does not explicitly crashes without the resample. Working on it now...

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/OpenEOProcessesSpec.scala
```diff
@@ -105,6 +109,62 @@ object OpenEOProcessesSpec {
...
+
+  def toIndex(key: SpaceTimeKey): BigInt = {
+    println("toIndex(" + key + ") returns: " + constant)
+    constant
+  }
+
+  def indexRanges(keyRange: (SpaceTimeKey, SpaceTimeKey)): Seq[(BigInt, BigInt)] = {
+    // Emile: Not sure what this function should do,
+    // but as there is only one partition that can get returned, this can't be wrong.
+    Seq((toIndex(keyRange._1), toIndex(keyRange._2)))
```
EmileSonneveld, line 151: @jdries Do you think this code is ok?

openeo-geotrellis/src/main/java/org/openeo/sparklisteners/GetInfoSparkListener.scala
```diff
@@ -29,21 +28,25 @@ class GetInfoSparkListener extends SparkListener {
     println("*****************************************************************")
   }
 
+  override def onApplicationEnd(applicationEnd: SparkListenerApplicationEnd): Unit = {
```
EmileSonneveld, line 31: I added logging, because CI gave an error in this branch. Only happened once, but I'll keep the log anyway for if it would happen in the future.

jdries: ok

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/OpenEOProcessesSpec.scala
jdries: I believe so!

Issue #90, opened 2022-10-28
[Performance] aggregate_temporal -> special case for SparseSpaceOnlyPartitioner


jdries:
In case SparseSpaceOnlyPartitioner is set on a cube, we immediately have full timeseries per partition. This allows doing aggregate_temporal in a map_partitions call, avoids a shuffle, and thus has potential to reduce shuffle output size.

jdries:
Implemented and seems to work, but how do we test this?

jdries:
For testing: can we record spark jobs and stages using:
 https://spark.apache.org/docs/latest/api/java/index.html?org/apache/spark/scheduler/SparkListener.html

We would then be able to put some asserts on how many stages are triggered by a job run?

PR #112, opened 2023-01-18
#90 Extend medianComposite() test to check if the partition optimisat…

EmileSonneveld:
…ion is working.


jdries:
looks like what I intended, I think we can merge

Issue #104, opened 2022-12-17
non-utm + non-aligned coords: x-aligned offset error


jdries:
When working in EPSG:3035, and requesting area that is not rounded to 10m, error happens.

EmileSonneveld:
Could reproduce by modifying `testCreoNonNativeProjection()`
`java.lang.IllegalArgumentException: requirement failed: x-aligned: offset by CellSize(10.0,10.0) 0.9`

EmileSonneveld:
Related: https://github.com/locationtech/geotrellis/issues/3292

PR #117, opened 2023-01-23
#104 roundExtendToGridOutwards to avoid EPSG:3035 issue.

EmileSonneveld:


jdries:
The main risk with this approach:
- the exact bounding box requested by the user may not be returned
- the input rasters may not align with the layout that is constructed here.

Up until now, we always ensured that at least one of these conditions is met. Especially when using a geographic crs, I would avoid this approach, as that would almost certainly break some workflows that rely on specific alignments.

In this specific case, the requested projection is different from input data projection, so it's impossible to align with input rasters. Hence the best we can do is satisfy the first condition?

EmileSonneveld:
@jdries This should be fine now. It is redoing another build for a change in `develop` branch I think.
For the GIFF equivalent, I logged a new issue: https://github.com/Open-EO/openeo-geotrellis-extensions/issues/120

Issue #109, opened 2023-01-15
upgrade to spark 3.3


jdries:

bossie:
Release notes for Spark 3.3.1: https://spark.apache.org/releases/spark-release-3-3-0.html

This build runs fine: https://jenkins.vgt.vito.be/job/openEO/job/openeo-geotrellis-extensions/job/109-upgrade-to-spark-33/

SLF4J should just pick up the Log4j 2.x that Spark provides; there should be no code changes in this case but we'll probably have to update its configuration: https://logging.apache.org/log4j/2.x/manual/configuration.html

The openeo-logging artifact currently builds against Spark 3.2's Log4j 1.x so that will bomb at runtime; Log4j 2.x has JSON support out of the box so we can probably get rid of this artifact altogether: https://logging.apache.org/log4j/2.x/manual/json-template-layout.html

Impact (of upgraded dependencies) on embedded tsservice: to be investigated.

bossie:
Next step: bumping the version and run openeo-geopyspark-driver tests against the new jar: https://github.com/Open-EO/openeo-geotrellis-extensions/blob/ee33ce2cd98e9b394863ab21a3f9093f27b1ef24/Jenkinsfile#L98

Future step: branch off our Docker image building repos and create a Docker image with the new Spark in it.

bossie:
As discussed: I believe it would have been possible to exclude this Spark's dependency on Log4j 2.x and include Log4j 1.x ourselves to avoid having to update logging configuration and JSON output but we'll take this chance to upgrade to Log4j 2.x instead.

bossie:
`org.openeo:geotrellis-extensions:2.3.0_2.12-spark331` has been published to Artifactory (in libs-release-public instead of libs-snapshot-public but ok).

openeo-geoypyspark-driver runs its tests in a [vito-docker.artifactory.vgt.vito.be/almalinux8.5-spark-py-openeo:3.2.0 container](https://github.com/Open-EO/openeo-geopyspark-driver/blob/f9d7f93e0d611c51d6242d7cc198d1cc3d86f6ce/Jenkinsfile#L27) against [the Spark 3.3.1 jars](https://github.com/Open-EO/openeo-geopyspark-driver/blob/f9d7f93e0d611c51d6242d7cc198d1cc3d86f6ce/scripts/get-jars.py#L65-L67).

bossie:
[...]

PR #124, opened 2023-02-06
109 upgrade to spark 33

bossie:
Implements #109.


Jenkinsfile
```diff
@@ -10,7 +10,7 @@ node ('devdmz') {
         sh "rm -rf .git/"
         checkout scm
         rel_version = getMavenVersion()
-        build()
+        build(tests = false) // FIXME: restore tests
```
bossie, line 13: Restore these tests.

PR #129, opened 2023-02-17
Custom asset names

EmileSonneveld:


EmileSonneveld:
Will already merge to make artifact that openeo-geopyspark-driver CI can use. If causing trouble, it can always get reverted.

PR #130, opened 2023-02-22
Use larger timeout for all Http calls.

### LINKED ISSUE #342: SocketTimeoutException: Read timed out towards SHub Catalog API
The cropsar process graph [mentioned](https://github.com/Open-EO/openeo-geopyspark-driver/issues/279#issuecomment-1429360601) in #279 will still not succeed after the changes in that ticket to avoid SHub batch processes: `java.net.SocketTimeoutException: Read timed out`.

Is this another manifestation of https://github.com/Open-EO/openeo-geotrellis-extensions/issues/93? It looks like there are integration tests that currently suffer from the same problem; is it a transient error or something that we have to fix?

To investigate why this happens.

EmileSonneveld:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/342

### DIFF
--- a/geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/AuthApi.scala
+++ b/geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/AuthApi.scala
@@ -43,5 +43,8 @@ class AuthApi {
         .valueOr(throw _)
     }
 
-  private def http(url: String): HttpRequest = Http(url).option(HttpOptions.followRedirects(true))
+  private def http(url: String): HttpRequest =
+    Http(url)
+      .option(HttpOptions.followRedirects(true))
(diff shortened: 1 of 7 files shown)


EmileSonneveld:
I used 40 seconds timeout, because that was already used in `ProcessApi.scala`

bossie:
Might as well include this one too: https://github.com/Open-EO/openeo-geotrellis-extensions/blob/a8d79e045feab92b327f53a95adc8d93995c00be/geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/MemoizedCuratorCachedAccessTokenWithAuthApiFallbackAuthorizerTest.scala#L17

bossie (changes requested):
Missed a call, otherwise fine.

bossie (approved):
Looks good.

PR #131, opened 2023-03-03
#128 Sentinelhub non-overlapping geometry

### LINKED ISSUE #128: Error: LINEARRING EMPTY when requesting data from SentinelHub
## Description

In one of our processes, we are fetching data from SentinelHub. However, we see something strange in this data retrieval step. When we request the data for a specific polygon in 2017, everything is working fine. When we request data for the same polygon in 2018, we get the following error message:
 
`Internal: Server error: Cannot create a polygon with exterior with fewer than 4 points: LINEARRING EMPTY (ref: r-02c11b2ef77243d98e9a09a5c2007aa4)`

Does anybody have an idea what could be causing this behavior? 
 
Some information to reproduce the issue:

## Python Sample
```python
dc = conn \
        .load_collection('SENTINEL2_L2A_SENTINELHUB',
                         bands=["B03", "B04", "B08", "sunAzimuthAngles", "sunZenithAngles", "viewAzimuthMean",
                                "viewZenithMean", 'SCL']) \
        .filter_spatial(features) \
        .filter_temporal(date_range)
```

## Processing Graph
```json
"loadcollection1": {
        "process_id": "load_collection",
        "arguments": {
            "bands": [
                "B03",
                "B04",
                "B08",
                "sunAzimuthAngles",
                "sunZenithAngles",
                "viewAzimuthMean",
                "viewZenithMean",
                "SCL"
            ],
            "id": "SENTINEL2_L2A_SENTINELHUB",
            "spatial_extent": null,
            "temporal_extent": null
        }
    },
    "filterspatial1": {
        "process_id": "filter_spatial",
        "arguments": {
            "data": {
                "from_node": "loadcollection1"
            },
            "geometries": {
                "type": "FeatureCollection",
                "features": [
                    {
                        "type": "Feature",
                        "properties": {
                            "id": 1
                        },
                        "geometry": {
                            "type": "Polygon",
                            "coordinates": [
                                [
                                    [
                                        -0.20057735970203852,
                                        53.24811471153172
                                    ],
                                    [
                                        -0.19884885496214083,
                                        53.24840102826777
                                    ],
                                    [
                                        -0.19791567448906724,
                                        53.24635439382115
                                    ],
                                    [
                                        -0.19964948139074376,
                                        53.246089285732204
                                    ],
                                    [
                                        -0.20057735970203852,
                                        53.24811471153172
                                    ]
                                ]
                            ]
                        }
                    }
                ]
            }
        }
    },
    "filtertemporal1": {
        "process_id": "filter_temporal",
        "arguments": {
            "data": {
                "from_node": "filterspatial1"
            },
            "extent": [
                "2018-01-01",
                "2018-12-31"
            ]
        },
        "result": true
    }
}
```

EmileSonneveld:
The problems was that the requested area was just touching the edge of a large sentinel tile. During CRS reprojection, the requested area did not intersect anymore, causing an error.

Now, the requested area get buffered/padded so that conversion errors are no problem anymore.

Aditionally, in the future, the code would not crash on some missing intersections, and will show better error message if no Feature has been found.

Note: SentinelHub also returns a Feature polygon in a different CRS. If We get more problems with this code, we can consider using this Polygon, as it is probably the one used internally. 
It is stored under `feature.properties["proj:geometry"]`

In the picture we see in brown the used Feature geometry. In green the unused geometry that get's passed too by SentinelHub and in yellow, the user input geometry:
image


EmileSonneveld:
Relevant commit in other repo: https://github.com/Open-EO/openeo-geopyspark-driver/commit/81e2763de2eadf61e3a0dbf458c5a792c5c66f97

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
```diff
@@ -334,10 +334,31 @@ class PyramidFactory(collectionId: String, datasetId: String, catalogApi: Catalo
 
             tracker.addInputProducts(collectionId, features.keys.toList.asJava)
 
-            val featureIntersections = for {
-              feature <- sc.parallelize(features.values.toSeq, math.max(1, features.size / 10))
-              reprojectedFeature = feature.reproject(LatLng, boundingBox.crs)
-            } yield Feature(reprojectedFeature intersection multiPolygon, reprojectedFeature.data.toLocalDate.atStartOfDay(UTC))
+            val featuresRDD = sc.parallelize(features.values.toSeq, math.max(1, features.size / 10))
```
bossie: A good opportunity to get rid of `math.max`: a Scala Int already has a `max` method so replace this with: `1 max (features.size / 10)`

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
```diff
@@ -334,10 +334,31 @@ class PyramidFactory(collectionId: String, datasetId: String, catalogApi: Catalo
...
+            val featureIntersections = featuresRDD.flatMap(feature => {
+              val reprojectedFeature = feature.reproject(LatLng, boundingBox.crs)
+              // In test over England, there where up to 0.003 deviations on long line segments due to curvature change between CRS
+              val multiPolygonBuffered = multiPolygon
+                .reproject(polygons_crs, LatLng)
+                .buffer(0.006)
+                .reproject(LatLng, polygons_crs)
```
bossie: Buffering `multiPolygon` is only necessary once so a first improvement is to move `multiPolygonBuffered` outside of the `flatMap`.

This step buffers `multiPolygon` with a constant 0.006°. Can you reproject the buffer distance of 0.006° to `polygons_crs` once (again, a constant) to avoid reprojecting `multiPolygon` twice?

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
```diff
@@ -334,10 +334,31 @@ class PyramidFactory(collectionId: String, datasetId: String, catalogApi: Catalo
...
+              val intersect = reprojectedFeature intersection multiPolygonBuffered
```
bossie: Rename this variable to `intersection`.

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
```diff
@@ -334,10 +334,31 @@ class PyramidFactory(collectionId: String, datasetId: String, catalogApi: Catalo
...
+
+              if (intersect.isEmpty) {
+                logger.debug(s"shub returned a Feature that actually does not intersect with our requested polygons.")
```
bossie: Log this particular feature as well.

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
```diff
@@ -334,10 +334,31 @@ class PyramidFactory(collectionId: String, datasetId: String, catalogApi: Catalo
...
+                Seq()
+              } else
+                Seq(Feature(intersect, reprojectedFeature.data.toLocalDate.atStartOfDay(UTC)))
```
bossie: You return a collection with 0 or 1 element(s): that's an `Option`. You can return `None` and `Some(Feature(...))` instead.

geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/PyramidFactoryTest.scala
```diff
@@ -527,7 +529,9 @@ class PyramidFactoryTest {
       from_date = ISO_OFFSET_DATE_TIME format date,
       to_date = ISO_OFFSET_DATE_TIME format date,
       band_names = Seq("HV", "HH").asJava,
-      metadata_properties = Collections.singletonMap("polarization", Collections.singletonMap("eq", "DH"))
+      metadata_properties = util.Collections.emptyMap[String, util.Map[String, Any]],
+      // gives no results with "polarization" filter?
+      //metadata_properties = Collections.singletonMap("polarization", Collections.singletonMap("eq", "DH")),
```
bossie: Possibly. These comments can be removed.

geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/PyramidFactoryTest.scala
```diff
@@ -1028,4 +1038,40 @@ class PyramidFactoryTest {
     verify(catalogApiSpy, atLeastOnce()).search(eqTo("sentinel-1-grd"), any(), eqTo(LatLng),
       eqTo(from), eqTo(ZonedDateTime.parse("2020-04-09T23:59:59.999999999Z")), any(), eqTo(util.Collections.emptyMap()))
   }
+
+  @Test
+  def testIssue128(): Unit = {
```
bossie: Needs a more descriptive name.

geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/PyramidFactoryTest.scala
```diff
@@ -1028,4 +1038,40 @@ class PyramidFactoryTest {
...
+    val endpoint = "https://services.sentinel-hub.com"
+
+    val catalogApiSpy = spy(new DefaultCatalogApi(endpoint))
```
bossie: Can be a regular `DefaultCatalogApi` instead of a spy.

geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/PyramidFactoryTest.scala
```diff
@@ -1028,4 +1038,40 @@ class PyramidFactoryTest {
...
+
+    val pyramidFactory = new PyramidFactory("sentinel-2-l2a", "sentinel-2-l2a", catalogApiSpy,
+      new DefaultProcessApi(endpoint), authorizer, sampleType = FLOAT32)
+
+    val polygons_crs = CRS.fromEpsgCode(32630)
+    val geojson = IOUtils.toString(getClass.getResource("/issue-128.geojson"))
```
bossie: `IOUtils.toString` is deprecated.

The geojson file needs a more descriptive name.

geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/PyramidFactoryTest.scala
```diff
@@ -1028,4 +1038,40 @@ class PyramidFactoryTest {
...
+    var multiPolygon = MultiPolygon(GeoJson.parse[Polygon](geojson))
```
bossie: Can be a val.

geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/PyramidFactoryTest.scala
```diff
@@ -1028,4 +1038,40 @@ class PyramidFactoryTest {
...
+    val date = "2018-10-07T00:00:00+00:00"
+    val sc: SparkContext = SparkUtils.createLocalSparkContext("local[*]", appName = getClass.getSimpleName)
+
+    try {
+      val datacubeParams = new DataCubeParameters()
+      datacubeParams.tileSize = 256
+      datacubeParams.layoutScheme = "FloatingLayoutScheme"
+      datacubeParams.partitionerIndexReduction = 7
+
+      // Should not throw error after #128 fix:
```
bossie: Which error?

geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/PyramidFactoryTest.scala
```diff
@@ -1028,4 +1038,40 @@ class PyramidFactoryTest {
...
+      val ret = pyramidFactory.datacube_seq(
+        polygons = Array(multiPolygon),
+        polygons_crs = polygons_crs,
+        from_date = date,
+        to_date = date,
+        band_names = util.Arrays.asList("B03"),
+        metadata_properties = util.Collections.emptyMap(),
+        datacubeParams,
```
bossie, line 1073: Never knew you could end your argument list with a comma in Scala. :thinking:

geotrellis-sentinelhub/pom.xml
```diff
@@ -70,16 +70,29 @@
...
+            <version>5.8.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <version>5.8.1</version>
             <scope>test</scope>
         </dependency>
         <dependency>
```
bossie, line 98: Are these dependencies used?

bossie (changes requested):
Added remarks inline.

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
EmileSonneveld: fixed

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
EmileSonneveld: True, reimplemented! Uses a bit more code now, but is probably way more performant.

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
EmileSonneveld: renamed

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
EmileSonneveld: done

geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/PyramidFactoryTest.scala
EmileSonneveld: true! Fixed

geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/PyramidFactoryTest.scala
EmileSonneveld: Instead of the `IOUtils.toString` workaround, would it be OK to move `ProjectedPolygons.fromVectorFile()` to `geotrellis-common` so that it can be used everywhere?

geojson renamed.

geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/PyramidFactoryTest.scala
EmileSonneveld: added comments

geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/PyramidFactoryTest.scala
EmileSonneveld: only when the arguments are multiline :)

geotrellis-sentinelhub/pom.xml
EmileSonneveld: They where used during debugging. I kept them as it keeps the dependencies in sync with `openeo-geotrellis/pom.xml`

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
EmileSonneveld: Did not know, changed!

geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/PyramidFactoryTest.scala
EmileSonneveld: Removed

geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/PyramidFactoryTest.scala
EmileSonneveld: renamed

geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/PyramidFactoryTest.scala
EmileSonneveld: Changed

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
bossie: I meant the geometry but I'll fix that.

geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/PyramidFactoryTest.scala
bossie: I'm not sure what the implications are of making submodules dependent on each other but the code was simple enough to replace.

EmileSonneveld: There are already some submodule dependencies, like `geotrellis-common` that is used on multiple places. But the solution you propose is way less impactful for sure!

EmileSonneveld: The `GeoJson.parse` solution does not parse the CRS from the `.geojson`. But as we know the file beforehand, and it is in `LatLng` (as the standard demands), that is safe enough.

PR #132, opened 2023-03-07
#70 Replace ID with URL in SentinelHub Metadata

### LINKED ISSUE #70: Report Sentinelhub features in metadata
For Terrascope collections, the batch job metadata now stores a link to the original products that were used to generate the result.
The 'derived-from' rel type should be used for this.

Terrascope doesn't use real links yet, but rather an identifier, this is also possible in the absence of decent links, e.g. :
urn:eop:VITO:TERRASCOPE_S2_TOC_V2:S2B_20200211T111109_30TUM_TOC_V200

The code in FileLayerProvider that does this looks like this:
`BatchJobMetadataTracker.tracker("").addInputProducts(openSearchCollectionId,overlappingFeatures.map(_.id).asJava)`

EmileSonneveld:


EmileSonneveld:
-> Testing this PR in dev deployment. I expect no big impact.

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/CatalogApi.scala
```diff
@@ -77,6 +80,50 @@ object DefaultCatalogApi {
...
+  private def getFeatureID(js: Json): String = {
+    val cursor = js.hcursor
+    val id = cursor.downField("id")
+    id.as[String] match {
+      case Right(i) => i
+      case _ =>
+        id.as[Int] match {
+          case Right(i) => i.toString
+          case _ => throw new RuntimeException("Feature expected to have \"ID\" field" + cursor.history)
+        }
```
bossie, line 95: Pattern matching on Either can be replaced with `valueOr(throw new ...)`.

But I have the feeling that this method is unnecessary because the feature IDs are already extracted (they are the keys in the `PagedJsonFeatureCollectionMap`).

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/CatalogApi.scala
```diff
@@ -174,21 +221,24 @@ class DefaultCatalogApi(endpoint: String) extends CatalogApi {
...
-            feature.mapData { properties =>
-              val Some(datetime) = for {
-                properties <- properties.asObject
-                json <- properties("datetime")
-                datetime <- json.asString
-              } yield ZonedDateTime.parse(datetime, ISO_OFFSET_DATE_TIME)
-
-              datetime
-            })
+        page.features.foreach { fJson =>
```
bossie, line 230: Prefer a `map` (or in this case, `mapValues` because you already have the feature ID) in favor of a `foreach` with a side-effect.

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/CatalogApi.scala
```diff
@@ -174,21 +221,24 @@ class DefaultCatalogApi(endpoint: String) extends CatalogApi {
...
+          type F = Feature[Geometry, Json]
+          fJson.as[F].foreach(feature => {
+            val key = getFeatureID(fJson)
+            val newFeature = feature.mapData { properties =>
+              val Some(dateTime) = getDateTime(properties)
+              val selfUrl = getSelfUrl(fJson)
```
bossie, line 236: `selfUrl` can be moved out of `mapData`.

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
```diff
@@ -332,12 +332,12 @@ class PyramidFactory(collectionId: String, datasetId: String, catalogApi: Catalo
                 from, atEndOfDay(to), accessToken, Criteria.toQueryProperties(metadata_properties))
             }
 
-            tracker.addInputProducts(collectionId, features.keys.toList.asJava)
+            tracker.addInputProducts(collectionId, features.map(p => p._2.data.selfUrl.getOrElse(p._1)).toList.asJava)
```
bossie, line 335: `features.map { case (id, Feature(_, FeatureData(_, selfUrl))) => selfUrl.getOrElse(id) }` is more readable than _1, _2 etc.

geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/MadeToMeasureCatalogApiTest.scala
```diff
@@ -24,9 +24,9 @@ class MadeToMeasureCatalogApiTest {
     val actualFeatures = catalogApi.search(collectionId, geometry, geometryCrs, from, to, accessToken, queryProperties)
 
     val expectedFeatures = Map(
-      "0" -> Feature(geometry, from),
-      "1" -> Feature(geometry, from plusDays 1),
-      "2" -> Feature(geometry, from plusDays 2)
+      "0" -> Feature(geometry, FeatureData(from, None)),
+      "1" -> Feature(geometry, FeatureData(from plusDays 1, None)),
+      "2" -> Feature(geometry, FeatureData(from plusDays 2, None))
```
bossie, line 29: You might want to give a name things like None, null, true, false etc to improve readability e.g.


```scala
val selfUrl = None
...
"0" -> Feature(geometry, FeatureData(from, selfUrl))
```

bossie (commented):
Some implementation remarks inline, otherwise fine.

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/CatalogApi.scala
EmileSonneveld: Fixed in https://github.com/Open-EO/openeo-geotrellis-extensions/pull/134

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/CatalogApi.scala
EmileSonneveld: `page.features` is a simple `List[JSON]`, so some parsing is needed to convert it to a `Map[String, Feature[...]]`
A fix was made in https://github.com/Open-EO/openeo-geotrellis-extensions/pull/134

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/CatalogApi.scala
EmileSonneveld: Moved in https://github.com/Open-EO/openeo-geotrellis-extensions/pull/134

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/PyramidFactory.scala
EmileSonneveld: Fixed in https://github.com/Open-EO/openeo-geotrellis-extensions/pull/134

geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/MadeToMeasureCatalogApiTest.scala
EmileSonneveld: Fixed in https://github.com/Open-EO/openeo-geotrellis-extensions/pull/134

geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/CatalogApi.scala
bossie: I see what you mean. I was under the assumption that we could still use `page.getAllFeatures`.

PR #140, opened 2023-03-27
Shub better buffer

EmileSonneveld:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/375

This MR adds a test for this issue, and rolls back a buffer.


jdries:
merged cause it's urgent

PR #146, opened 2023-04-20
Auto polarization

EmileSonneveld:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/350


jdries:
fine for me, will leave some time for others to review but otherwise can be merged

PR #155, opened 2023-05-08
Sentinel offset

### LINKED ISSUE #144: apply Sentinel-2 scale + offset for new products
Sentinel-2 L2A products processed with latest version have a scale and offset that needs to be applied, otherwise users will get unexpected results.

> Please note that starting with 25th January 2022 (PB 04.00), to allow encoding negative surface reflectances that may occur over very dark surfaces, the dynamic range of Level-2A products is shifted by a band-dependent constant BOA_ADD_OFFSET. The value of this offset is available in the product metadata.

The default offset seems to be -1000 for all bands, but need to check product baseline.

EmileSonneveld:


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
```diff
@@ -59,15 +64,49 @@ class BandCompositeRasterSource(override val sources: NonEmptyList[RasterSource]
...
+      band match {
+        case b: geotrellis.raster.UShortArrayTile =>
+          geotrellis.raster.ShortConstantNoDataArrayTile(b.array.map(x =>
+            if (x == ushortNODATA) shortNODATA else (x + pixelValueOffset).toShort), band.cols, band.rows)
+        case b: geotrellis.raster.UByteArrayTile =>
+          geotrellis.raster.ByteConstantNoDataArrayTile(b.array.map(x =>
+            if (x == ubyteNODATA) byteNODATA else (x + pixelValueOffset).toByte), band.cols, band.rows)
+        case _ =>
+          // Not sure how to handle user defined nodata for example
+          throw new IllegalArgumentException("Can not yet combine 'pixelValueOffset' and '" + band.getClass.getName + "'. See more: https://github.com/Open-EO/openeo-geotrellis-extensions/issues/144")
```
EmileSonneveld: @bossie , ok to refer to ticket here?

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/layers/FileLayerProviderTest.scala
```diff
@@ -981,4 +1000,98 @@ class FileLayerProviderTest {
...
+    for (rest <- Seq(
+      "/eodata/Sentinel-2/MSI/L2A/2023/01/17/S2B_MSIL2A_20230117T104259_N0509_R008_T31UGS_20230117T120337.SAFE/manifest.safe",
+      "/eodata/Sentinel-2/MSI/L2A/2023/01/17/S2B_MSIL2A_20230117T104259_N0509_R008_T31UGS_20230117T120337.SAFE/GRANULE/L2A_T31UGS_A030636_20230117T104258/IMG_DATA/R10m/T31UGS_20230117T104259_B04_10m.jp2",
+      "/eodata/Sentinel-2/MSI/L2A/2023/04/05/S2A_MSIL2A_20230405T105031_N0509_R051_T31UFS_20230405T162253.SAFE/manifest.safe",
+      "/eodata/Sentinel-2/MSI/L2A/2023/04/05/S2A_MSIL2A_20230405T105031_N0509_R051_T31UFS_20230405T162253.SAFE/GRANULE/L2A_T31UFS_A040660_20230405T105026/IMG_DATA/R10m/T31UFS_20230405T105031_B04_10m.jp2",
+    )) {
+      val jp2File = new File(basePath, rest)
+      if (!jp2File.exists()) {
+        println("Copy from artifactory to: " + jp2File)
+        FileUtils.copyURLToFile(new URL(basePathArtifactory + rest), jp2File)
```
EmileSonneveld, line 1016: This function is runned 2x, totalling 4min in the test suite.  
@bossie , do you think there is an easy way to make this download faster? By cropping or lossy compressing the jp2 for example

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
```diff
@@ -59,15 +64,49 @@ class BandCompositeRasterSource(override val sources: NonEmptyList[RasterSource]
     }
 
   }
-  override def cellType: CellType = sources.map(_.cellType).reduceLeft(_ union _)
+
+  override def cellType: CellType = {
+    var _cellType = sources.map(_.cellType).reduceLeft(_ union _)
```
bossie: Favor `val`s instead of `var`s. Just call this one e.g. `val commonCellType = ...` and return
```scala
if (pixelValueOffset < 0) commonCellType match {
 // ...
} else commonCellType
```

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
```diff
@@ -59,15 +64,49 @@ class BandCompositeRasterSource(override val sources: NonEmptyList[RasterSource]
...
+    if (pixelValueOffset < 0) {
+      // Big chance the value will go under 0, so type needs to be signed
+      _cellType = _cellType match {
+        case UByteCellType => ByteCellType
+        case UByteConstantNoDataCellType => ByteConstantNoDataCellType
+        case UShortCellType => ShortCellType
+        case UShortConstantNoDataCellType => ShortConstantNoDataCellType
+        case default => default // Was probably already signed
```
bossie: Or just: `case _ => commonCellType`

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
```diff
@@ -59,15 +64,49 @@ class BandCompositeRasterSource(override val sources: NonEmptyList[RasterSource]
...
+      }
+    }
+    _cellType
+  }
 
   override def name: SourceName = sources.head.name
   override def bandCount: Int = sources.size
 
+  private def offsetBand(band: Tile): Tile = {
```
bossie: Naming is confusing, see below for an alternative.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
bossie: Looks brittle. Can't this just be replaced with (untested):

```scala
  private def withOffset(bandTile: Tile): Tile = {
    if (pixelValueOffset == 0) bandTile
    else if (cellType.isFloatingPoint) bandTile.convert(cellType).mapDouble(x => x + pixelValueOffset)
    else bandTile.convert(cellType).map(i => i + pixelValueOffset.toInt)
  }
```

bossie: Is this for the user or the developer? If a user bumps into this we're going to get notified anyway.

What you should do in any case is add the issue URL to your commit messages so Github can match commits with tickets.

If the commit is in the same repo as the ticket, you can use a shorthand, e.g. `#144` instead of `https://github.com/Open-EO/openeo-geotrellis-extensions/issues/144`

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
```diff
@@ -59,15 +64,49 @@ class BandCompositeRasterSource(override val sources: NonEmptyList[RasterSource]
...
+
+  override def cellType: CellType = {
+    var _cellType = sources.map(_.cellType).reduceLeft(_ union _)
+    if (pixelValueOffset < 0) {
+      // Big chance the value will go under 0, so type needs to be signed
+      _cellType = _cellType match {
+        case UByteCellType => ByteCellType
+        case UByteConstantNoDataCellType => ByteConstantNoDataCellType
+        case UShortCellType => ShortCellType
+        case UShortConstantNoDataCellType => ShortConstantNoDataCellType
```
bossie: Don't we need to support additional cell types, esp. with `pixelValueOffset` being a `Double`? Maybe you could extract this into a function e.g. `toSigned(cellType)` if there isn't one already.

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/layers/FileLayerProviderTest.scala
```diff
@@ -40,7 +43,7 @@ object FileLayerProviderTest {
   // - BeforeClass
   // - AfterClass
   //
-  // This order feels arbitrary, so I made the code robust against order changes.
+  // This order is because of multiple Test framework versions, so I made the code robust against order changes.
```
bossie: If the problem is that we're mixing JUnit 4 and 5 APIs and they don't play well together I would just get rid of the former and replace @BeforeClass with of @BeforeAll etc.

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/layers/FileLayerProviderTest.scala
```diff
@@ -362,6 +371,16 @@ class FileLayerProviderTest {
...
 
+  class MockOpenSearchFeatures(mockedFeatures:Array[OpenSearchResponses.Feature]) extends OpenSearchClient {
+    override def getProducts(collectionId: String, dateRange: Option[(ZonedDateTime, ZonedDateTime)], bbox: ProjectedExtent, attributeValues: collection.Map[String, Any], correlationId: String, processingLevel: String): Seq[OpenSearchResponses.Feature] = {
+      mockedFeatures
+    }
+
+    override protected def getProductsFromPage(collectionId: String, dateRange: Option[(ZonedDateTime, ZonedDateTime)], bbox: ProjectedExtent, attributeValues: collection.Map[String, Any], correlationId: String, processingLevel: String, startIndex: Int): OpenSearchResponses.FeatureCollection = ???
+
+    override def getCollections(correlationId: String): Seq[OpenSearchResponses.Feature] = ???
+  }
```
bossie: You can write the existing `MockOpenSearch` in terms of this `MockOpenSearchFeatures` to avoid code duplication.

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/layers/FileLayerProviderTest.scala
bossie: Those are some good options but they still require some effort.

BTW I used test files in the past but stored them on NFS instead of in Artifactory; I don't think there's much to be gained there though.

https://github.com/Open-EO/openeo-geotrellis-extensions/blob/68143219e6882dffd0d05a46aaf29e530c0b93a5/geotrellis-sentinelhub/src/test/scala/org/openeo/geotrellissentinelhub/PyramidFactoryTest.scala#L325C4-L326

Ideally you generate the jp2 yourself in code with exactly the pixels that you need for the test and avoid the download altogether.

bossie (changes requested):
Added remarks inline. Have you compared the results against those on Terrascope?

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
EmileSonneveld: When using the convert, the nodata values go wrong. The source tiles indicate that 0 would be the nodata value, but when loading, a far negative value is used. @bossie , do you know if the code should read the nodata value out of the metadata?

EmileSonneveld: With some fiddling, that snippet works perfectly. Thanks!

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/layers/FileLayerProviderTest.scala
EmileSonneveld: done

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
EmileSonneveld: True, I forget to label some commits. I'll see if I can put a pre-commit hook to remind me

EmileSonneveld: Also, I removed the url from the message

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/layers/FileLayerProviderTest.scala
EmileSonneveld: I'll leave as it is then. I tried to refer to them by URL, hoping that they would only be partially downloaded, but I got gdal related errors and aborted that plan

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
EmileSonneveld: done

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
EmileSonneveld: I made all types explicit in `toSigned` and made it throw an error when an unknown type would apear.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -89,4 +90,32 @@ package object geotrellis {
 
     Region.of(regionName)
   }
+
+  def toSigned(cellType: CellType): CellType = {
+    cellType match {
+      case UByteCellType => ByteCellType
+      case UByteConstantNoDataCellType => ByteConstantNoDataCellType
+      case c: UByteUserDefinedNoDataCellType => ByteUserDefinedNoDataCellType(c.noDataValue)
```
bossie: No need to introduce a name here, just leverage pattern matching:

```scala
case UByteUserDefinedNoDataCellType(noDataValue) => ByteUserDefinedNoDataCellType(noDataValue)
```

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -89,4 +90,32 @@ package object geotrellis {
...
+      case _: ShortUserDefinedNoDataCellType => cellType
+      case IntConstantNoDataCellType => cellType
+      case IntCellType => cellType
+      case _: IntUserDefinedNoDataCellType => cellType
+      case FloatCellType => cellType
+      case _: FloatUserDefinedNoDataCellType => cellType
+      case DoubleConstantNoDataCellType => cellType
+      case DoubleCellType => cellType
+      case _: DoubleUserDefinedNoDataCellType => cellType
+      case _ => throw new IllegalArgumentException("Can convert to unsigned equivalent: '" + cellType.getClass.getName + "'.")
```
bossie: Typo: Cannot?

String interpolation makes it easier to read:

```scala
s"Cannot ... '${cellType.getClass.getName}'."
```

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -89,4 +90,32 @@ package object geotrellis {
...
+      case c: UByteUserDefinedNoDataCellType => ByteUserDefinedNoDataCellType(c.noDataValue)
+      case UShortCellType => ShortCellType
+      case UShortConstantNoDataCellType => ShortConstantNoDataCellType
+      case c: UShortUserDefinedNoDataCellType => ShortUserDefinedNoDataCellType(c.noDataValue)
+      case FloatConstantNoDataCellType => cellType
+      case ShortConstantNoDataCellType => cellType
+      case BitCellType => cellType
+      case ByteConstantNoDataCellType => cellType
+      case ByteCellType => cellType
+      case _: ByteUserDefinedNoDataCellType => cellType
```
bossie: Or to align with the rest:

```scala
case ByteUserDefinedNoDataCellType(_) => cellType
```

bossie (changes requested):
Some aesthetic remarks, otherwise fine.

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/layers/FileLayerProviderTest.scala
EmileSonneveld: File is converted to only use JUnit 5 classes. Other files around the project have still JUnit 4 classes, but it is easier to convert them gradually when some other fix is made there.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
EmileSonneveld: Nice syntax. Changed

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
EmileSonneveld: Indeed, fixed!

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
EmileSonneveld: Changed too. And added `ByteUserDefinedNoDataCellType` to the `toSigned` test

EmileSonneveld:
Tweaks applied. And CI fixed

bossie (approved):

PR #158, opened 2023-05-11
Add array_apply.

### LINKED ISSUE #154: implement array_apply
Add to OpenEOProcessScriptBuilder https://processes.openeo.org/#array_apply

EmileSonneveld:


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcessScriptBuilder.scala
```diff
@@ -906,6 +907,20 @@ class OpenEOProcessScriptBuilder {
...
 
+  private def arrayApplyFunction(arguments: java.util.Map[String, Object]): OpenEOProcess = {
+    logger.warn("array_apply under development! Don't use yet")
+    val storedArgs = contextStack.head
+    val inputFunction = storedArgs("data")
+    val processFunction = storedArgs("process")
+
+    val bandFunction = (context: Map[String, Any]) => (tiles: Seq[Tile]) => {
+      val data: Seq[Tile] = evaluateToTiles(inputFunction, context, tiles)
+      val mappedValues: Seq[Tile] = evaluateToTiles(processFunction, context, data)
```
jdries, line 918: not entirely sure, but here I would expect some kind of loop over data?
Unless if the loop is somehow part of processFunction already?
Of course, when thinking about it, there's quite a few functions that will anyway loop over the sequence and apply to each tile individually.

EmileSonneveld: I still need to check if it goes over every band and time correctly.  But for a single tile it works

EmileSonneveld:
It was easier for me to test this on `openeo-dev.vito.be`.
Array apply gives the same result netCDF as calling a function in the original way:
```python
# the returning datacube has multiple bands and multiple times
datacube = connection.load_collection(
    "TERRASCOPE_S2_TOC_V2",
    spatial_extent={"west": 5.07, "south": 51.215, "east": 5.08, "north": 51.22},
    temporal_extent=["2020-03-01", "2020-03-30"],
    bands=["B01", "B02", "B03"],
)
# process = lambda x: cos(x)  # <- baseline
process = lambda x: x.array_apply(cos)
datacube_applied = datacube.apply_dimension(dimension='t', process=process)
job = datacube_applied.execute_batch(
    title=os.path.basename(__file__),
    format="netCDF",
)
```
@jdries Is this enough for a test?
image

jdries:
yes, I assume we also have unit tests right? Then we can merge, and we can do a more advanced case with rank composites.

EmileSonneveld:
@jdries I added unit tests JVM side and Python side:
https://github.com/Open-EO/openeo-geotrellis-extensions/commit/1876e806c73871505212e4add44712c687a7ab04 and https://github.com/Open-EO/openeo-geopyspark-driver/commit/1387884b14cee8160b8776c050ecd1b1522cc558

For this ticket I worked the main branches because I only saw how to test datacubes with multiple dates once deployed.

PR #165, opened 2023-05-23
Move error test from Python to Scala.

### LINKED ISSUE #429: job and user id in conftest
https://github.com/Open-EO/openeo-geopyspark-driver/blob/b5ea93ed110dcfb433f117fb20de93ded7831324/tests/conftest.py#L134-L135

@EmileSonneveld  why did you add these env vars here in conftest? 
These values don't seem to be used/checked elsewhere

EmileSonneveld:
This also removes an unused function in LogErrorSparkListener


openeo-logging/src/test/scala/org/openeo/logging/OpenEOBatchJobJsonLogLayoutTest.scala
```diff
@@ -4,19 +4,41 @@ import org.apache.logging.log4j.core.LoggerContext
...
+  private lazy val temporaryFolder: String = {
+    val _temporaryFolder = new File("tmp/" + UUID.randomUUID().toString).getAbsolutePath
+    Files.createDirectories(Paths.get(_temporaryFolder))
+    _temporaryFolder
+  }
 
   @BeforeClass
-  def initializeLog4j(): Unit = loggerContext = Configurator.initialize(null, "classpath:log4j2-batch.xml")
+  def initializeLog4j(): Unit = loggerContext = {
+    // Do a hard replacement of variables in the XML to avoid influence of thread context
```
EmileSonneveld: @bossie Before, logs from different threads would not have the 'logFile' defined, and would write to a file called "${ctx:logFile}". This fix is a bit ugly, but quite stable I think. Other solutions could be:
- Use an environment variable, but that is difficult to change from scala
- Use static path as log location, but then the file needs to be cleared before each test.

Maybe worth changing "${ctx:logFile}" to a different token? Or do you think this is ok for this test case?

bossie: I think this is easier: https://github.com/Open-EO/openeo-geotrellis-extensions/pull/167

EmileSonneveld: but I need the logs from all threads. #167 throws them away, right?

bossie: Right, it's the `TaskSetManager` logs that you want.

I updated #167 with a JUnit Rule for environment variables. Does this look better?

Please give your test a more descriptive name (I accidentally added it to #167 as well).

bossie (approved):
Looks good.

EmileSonneveld:
The tests that fail here are the same as on the develop branch:
- org.openeo.geotrellissentinelhub.BatchProcessingApiTest.getBatchProcess
- org.openeo.geotrellissentinelhub.BatchProcessingServiceTest.getBatchProcessStatus
- org.openeo.geotrellissentinelhub.BatchProcessingServiceTest.getBatchProcess

PR #167, opened 2023-05-23
ctx:logFile empty

bossie:
Logs from Spark UI etc. were written to a file actually called ${ctx:logFile}.

https://github.com/Open-EO/openeo-geopyspark-driver/issues/429


bossie:
Suggestions were applied in #165.

Issue #152, opened 2023-05-03
support bands with 10m and 20m resolution in single datacube (Sentinel-2)


jdries:
Currently, a MultibandTile consists of tiles with the same resolution.
Sentinel-2 however has bands in 10m and 20m resolution. 
We have 2 cases to try and keep bands in 20m:
- super-resolution use case requiring bands in original format
- reduce memory use & increase processing performance

jdries:
Proposed solution: create Tile implementation that looks like 10m, but uses downsampled array for storage.

JeroenVerstraelen:
The feature is live on the development servers and can be tested using:
```
connection: Connection = openeo.connect("https://openeo-dev.vito.be/openeo/1.0").authenticate_oidc()
sentinel2_data_cube: DataCube = connection.load_collection("TERRASCOPE_S2_TOC_V2",
    temporal_extent = ["2018-05-06", "2018-05-06"], bands = ["B02", "B05", "B08"], spatial_extent = {
        'west': 5.15183687210083,
        'east': 5.153381824493408,
        'south': 51.18192559252128,
        'north': 51.18469636040683,
        'crs': "EPSG:4326"
    })
sentinel2_data_cube.result_node().update_arguments(featureflags={'no_resample_on_read': True})
```

There are still some issue with the gridbounds calculations.
```
openeo.rest.OpenEoApiError: [500] Internal: Server error: Exception during Spark execution: GridBounds(0,0,-1,15) do not intersect 7x16 (ref: r-d12ea993bd844226b32901e1a229be85)
```

JeroenVerstraelen:
The issue has been solved and tested on openeo-dev.

PR #168, opened 2023-05-24
#152 add support for ResampledTiles

JeroenVerstraelen:
Added noResampleOnRead datacube parameter
Added ResampledRasterSource
Added ResampledTile
Added testS2UpsampledTilesFromLayer

There is one small issue with CroppedTile.convert, this creates a new ArrayTile and destroys the underlying ResampledTile. We use this whenever we read from a BandCompositeRasterSource. To circumvent this issue I introduced the ResampledTile.cropAndConvert method.


jdries:
looks generally fine, and safe to merge. Maybe we can merge these things into develop already, but two things to consider:
- ResampleTile was added to tile_grid package, not sure if that is a logical place, maybe consider putting it somewhere under geotrellis-common?
- Add unit test on the new tile and RasterSource? Maybe we can even copy basic tests from geotrellis itself?

PR #170, opened 2023-06-15
Select tile size

### LINKED ISSUE #169: default tilesize tuning for synchronous jobs: select larger tile size faster
testing shows that larger tile sizes are faster to load, especially on creodias.
Let's reduce the threshold for switching from 256 to 512.

EmileSonneveld:
This fixes some test failures that where introduced here: https://github.com/Open-EO/openeo-geotrellis-extensions/commit/07f460963707a8ddaebc6c5dd35ff73e69d5f73a
By taking into account the tile size, and avoiding a value compare at the edge of an image. 

Integration tests here might still fail after this PR, but those get handles by [an other PR](https://github.com/Open-EO/openeo-geotrellis-extensions/pull/171).

### DIFF
--- a/openeo-geotrellis/src/test/scala/org/openeo/geotrellis/layers/FileLayerProviderTest.scala
+++ b/openeo-geotrellis/src/test/scala/org/openeo/geotrellis/layers/FileLayerProviderTest.scala
@@ -190,24 +190,24 @@ class FileLayerProviderTest {
     val polygons1 = MultiPolygon(bbox1.extent.toPolygon())
     val (rasterSources1, metadata1) = _getSentinel5PRasterSources(bbox1, date, zoom)
     val sparseBaseLayer = FileLayerProvider.readMultibandTileLayer(rasterSources1, metadata1, Array(polygons1),
-                                                                   bbox1.crs, sc, retainNoDataTiles = false,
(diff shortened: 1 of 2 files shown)


openeo-geotrellis/src/test/scala/org/openeo/geotrellis/layers/FileLayerProviderTest.scala
```diff
@@ -804,8 +811,8 @@ class FileLayerProviderTest {
     assertEquals(0,minKey.col)
     assertEquals(0,minKey.row)
     assertEquals(crs,result._2.crs)
-    assertEquals(12,all.length)
-    //assertEquals((cols*rows).toInt,all.length)
+    assertTrue(1 <= all.length) // dependant on selectLayoutScheme
```
jdries, line 814: not sure if we really want to relax the assertion? It was in place to be really sure that we get the right amount of tiles back.

EmileSonneveld: I added a more strict assertion just below. I left the very relax assertion just in case.

jdries:
can merge!

PR #171, opened 2023-06-15
Update CollectionTests expected results.

EmileSonneveld:


JeroenVerstraelen (approved):
Looks good, the reference data changes appear to be very small and they're in line with the changes in the openeo-integrationtests repository.

PR #175, opened 2023-06-19
S2 offset

### LINKED ISSUE #172: apply Sentinel-2 BOA_ADD_OFFSET / RADIO_ADD_OFFSET
https://sentinels.copernicus.eu/web/sentinel/technical-guides/sentinel-2-msi/processing-baseline

Now the offset is hardcoded for all bands: https://github.com/Open-EO/openeo-geotrellis-extensions/issues/144
The offset should be read from the metadata xml and applied on only the relevant bands.

EmileSonneveld:
Now band specific. Using metadata.

### DIFF
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/ValueOffsetRasterSource.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/ValueOffsetRasterSource.scala
@@ -60,11 +60,13 @@ class ValueOffsetRasterSource(protected val rasterSource: RasterSource,
   override def metadata: RasterMetadata = rasterSource.metadata
 
   override protected def reprojection(targetCRS: CRS, resampleTarget: ResampleTarget, method: ResampleMethod, strategy: OverviewStrategy): RasterSource = {
-    rasterSource.reproject(targetCRS, resampleTarget, method, strategy)
+    val rs = rasterSource.reproject(targetCRS, resampleTarget, method, strategy)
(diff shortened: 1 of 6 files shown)


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/ValueOffsetRasterSource.scala
```diff
@@ -0,0 +1,90 @@
...
+      toSigned(commonCellType)
+    } else {
+      commonCellType
+    }
+  }
+
+  override def metadata: RasterMetadata = rasterSource.metadata
+
+  override protected def reprojection(targetCRS: CRS, resampleTarget: ResampleTarget, method: ResampleMethod, strategy: OverviewStrategy): RasterSource = {
+    rasterSource.reproject(targetCRS, resampleTarget, method, strategy)
```
JeroenVerstraelen: Note that when someone uses ValueOffsetRasterSource.reprojection, resample, convert then you don't return a ValueOffsetRasterSource. It's the best to wrap the output of these methods.

JeroenVerstraelen (changes requested):

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/ValueOffsetRasterSource.scala
EmileSonneveld: Good catch! I wrapped them and added tests for it.

JeroenVerstraelen (approved):

Issue #177, opened 2023-06-22
Sentinel2 L1C: SAA SZA VAA VZA bands not found


EmileSonneveld:
Probably here, changing finder.creodias.eu to zipper.creodias.eu could solve the problem.
Also filtering out the prefixes could be nicer.

https://github.com/Open-EO/openeo-geotrellis-extensions/blob/68143219e6882dffd0d05a46aaf29e530c0b93a5/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala#L884

Example error from kibana:
```
java.net.MalformedURLException: no protocol: /eodata/Sentinel-2/MSI/L1C/2023/05/30/S2B_MSIL1C_20230530T104629_N0509_R051_T31UFS_20230530T125255.SAFE/GRANULE/L1C_T31UFS_A032538_20230530T104623/MTD_TL.xml
	at java.base/java.net.URL.<init>(URL.java:645)
	at java.base/java.net.URL.<init>(URL.java:541)
	at java.base/java.net.URL.<init>(URL.java:488)
	at org.openeo.geotrellis.layers.FileLayerProvider.rasterSource$1(FileLayerProvider.scala:947)
	at org.openeo.geotrellis.layers.FileLayerProvider.$anonfun$deriveRasterSources$21(FileLayerProvider.scala:1011)
```

The error is about the prefix, but even without the good prefix, the url gives a 404: https://finder.creodias.eu/files/Sentinel-2/MSI/L1C/2023/05/30/S2B_MSIL1C_20230530T104629_N0509_R051_T31UFS_20230530T125255.SAFE/GRANULE/L1C_T31UFS_A032538_20230530T104623/MTD_TL.xml

jdries:
also check state of these bands SENTINEL2_L2A

EmileSonneveld:
This works now on CDSE and tests are added as to `test_collections.py` https://git.vito.be/projects/TPT/repos/os_creodias_openeo_k8s/browse/kube_resources/applications/integrationtests/tests/test_collections.py?at=797c76d2003343dbd42427fa8591c2777a1219ab#120-121

Also tested with example from Pratichhya: [pratichhya_angle_bands.json.txt](https://github.com/Open-EO/openeo-geotrellis-extensions/files/12596970/pratichhya_angle_bands.json.txt)

However, on `openeo.vito.be` this does not work atm.

EmileSonneveld:
[...]

PR #178, opened 2023-06-22
Fix xml band again

EmileSonneveld:
https://github.com/Open-EO/openeo-geotrellis-extensions/issues/177


EmileSonneveld:
@jdries here, FileLayerProvider should be more robust for URLs with various prefixes. I also used `zipper.creodias.eu` instead of `finder.creodias.eu`. Because the zipper works for me locally, and most tests are done with that one.

jdries:
approved, can merge

PR #197, opened 2023-08-22
support udf's that change spatial resolution of the datacube

### LINKED ISSUE #412: allow UDF's with xarray DataSet as input, with variables with different resolution
For super resolution use cases, some UDF's want to work on bands with mixed resolution: 10m and 20m.
Currently, UDF's work only on a DataArray, so resolution have to be aligned. If we define a new UDF signature that supports DataSet, we can also pass on bands with different resolutions.

If the input datacube is sentinel-2 with a mix of 10m and 20m bands, and a UDF with this signature is provided, we want to pass on bands in their original resolution.

This forum post from EvoLand has an extra twist: generating 5m data based on 10m and 20m inputs:
https://discuss.eodc.eu/t/super-resolution-with-udf-wrong-georeferencing-of-the-output-image/588

jdries:
Allow a UDF that transforms datacube spatial dimensions, to support cases like custom upsampling of the data.
A new udf signature is introduced that transforms the metadata itself. This allows the backend to determine updated cube metadata, without running the actual UDF.

https://github.com/Open-EO/openeo-geopyspark-driver/issues/412

### DIFF
--- a/openeo-geotrellis/pom.xml
+++ b/openeo-geotrellis/pom.xml
@@ -190,7 +190,7 @@
         <dependency>
             <groupId>black.ninia</groupId>
             <artifactId>jep</artifactId>
-            <version>3.9.1</version>
+            <version>4.1.1</version>
         </dependency>
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/udf/Udf.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/udf/Udf.scala
@@ -295,7 +295,7 @@ object Udf {
     // and then slice it!
 
     val newLayout = {
-      if (code.contains("convert_dimensions")) {
+      if (code.contains("apply_metadata")) {
         val newResolution = 5.0 //TODO determine based on convert_dimensions
         val crsCode = layer.metadata.crs.epsgCode.get
         val stepSize = layer.metadata.layout.cellSize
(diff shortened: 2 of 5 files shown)


openeo-geotrellis/pom.xml
```diff
@@ -190,7 +190,7 @@
         <dependency>
             <groupId>black.ninia</groupId>
             <artifactId>jep</artifactId>
-            <version>3.9.1</version>
+            <version>4.1.1</version>
```
JeroenVerstraelen, line 193: When reading the changelogs for jep I did come across this message for version 4.0.0:
> This is a major version increment and it includes removal of deprecated functionality and several changes making it incompatible with previous versions of Jep. This release also removes support for Python 2.

I will have to double check if there was a specific reason for why version 3.9.1 was chosen, but our use of Jep bindings is minimal so as long as all UDF tests work it is likely that we'll be ok. I will do some extra tests and go over the API changes just in case.

JeroenVerstraelen: Note that the jep version is also defined in these places:
https://github.com/Open-EO/openeo-geopyspark-driver/blob/master/setup.py#L82
https://github.com/Open-EO/openeo-geotrellis-extensions/blob/master/geotrellis-integrationtests/pom.xml#L174

JeroenVerstraelen: I checked the changes for the Jep library between versions 3.9.1 and 4.1.1: 

- The only important change is that setRedirectOutputStreams (whether to redirect Python streams to Java) has been split into redirectStdout, redirectStdErr. (https://github.com/ninia/jep/commit/f67b874b637e24ac3b725f2e89a57b51cded95e1) 
  - This is required to see the UDF logs.
- SharedInterpreters now also have an extra feature but it should not be relevant to our implementation:
[SharedInterpreters now share MemoryManager and java PyObjects. (#421)](https://github.com/ninia/jep/commit/2f6948ac19469b55d4800169a20d21ee418a8658)

I also tested this java build in the geopyspark-driver and it works without issues.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/udf/Udf.scala
```diff
@@ -35,7 +36,7 @@ object Udf {
   private def _createSharedInterpreter(): SharedInterpreter = {
     if (!isInterpreterInitialized) {
       val config = new JepConfig()
-      config.setRedirectOutputStreams(true)
+      //config.setRedirectOutputStreams(true)
```
JeroenVerstraelen: We should use the new redirectStdout, redirectStdErr methods here to avoid losing UDF logging.

JeroenVerstraelen (changes requested):

PR #220, opened 2023-10-23
Fix bandcount

### LINKED ISSUE #201: bug: merge_cubes without linear interpolation
Error message: The number of bands in the metadata 12/1 does not match the actual band count in the cubes (left/right): 2/1. You can fix this by explicitly specifying correct band labels.'

Repository: https://git.vito.be/projects/NCA/repos/extentmapping/browse?at=refs%2Fheads%2Ftests
 

Example notebook: https://git.vito.be/projects/NCA/repos/extentmapping/browse/notebooks/retrieve_S1_S2_timeseries_via_openeo.ipynb?at=refs%2Fheads%2Ftests 

The example notebook directly replicates the error – or see last job: j-3e198f201d7b49028bd665b41e85256e

The 12 band cube originates from merging Sentinel-2 and Sentinel-1. It seems like areas without Sentinel-2 data still are not properly filled with nodata bands somehow.

### LINKED ISSUE #217: aggregate_temporal_period: Band count must be greater than 0
Running `aggregate_temporal_period` on a datacube that contains a whole month without data can trigger an error. In this case, a `filter_spatial` should come after the aggregate node to trigger the error. Reproduced on `openeo-dev.vito.be` and `openeo.cloud`.
I would expect it to generate tiles with nodata instead of throwing an error.

Example in this attachment. Change the `temporal_extent` to enable or disable the error:  [aggregate_temporal_period.py.txt](https://github.com/Open-EO/openeo-geotrellis-extensions/files/12928983/aggregate_temporal_period.py.txt)

EmileSonneveld:
Summary:
- Some EmptyMultibandTiles where made with 0 bands. While there should be 'n' bands
- When asking EmptyMultibandTile for a list of bands, an empty array was returned, instead of a list of 'n' empty bands.

This caused problems when combining merge_cubes and aggregate_temporal.
A test has been added that goes over many combinations of those, to filter out bugs.

### DIFF
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/EmptyMultibandTile.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/EmptyMultibandTile.scala
@@ -37,7 +37,7 @@ class EmptyMultibandTile(val cols:Int, val rows:Int, val cellType:CellType, val 
 
   override def band(bandIndex: Int): Tile = EmptyMultibandTile.empty(cellType, cols, rows)
 
-  override def bands: Vector[Tile] = List.fill(bandCount)(EmptyMultibandTile.empty(cellType, cols, rows)).toVector
+  override def bands: Vector[Tile] = Vector.fill(bandCount)(EmptyMultibandTile.empty(cellType, cols, rows))
 
   override def subsetBands(bandSequence: Seq[Int]): MultibandTile = new EmptyMultibandTile(cols,rows,cellType,bandSequence.size)
 
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/EmptyMultibandTile.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/EmptyMultibandTile.scala
@@ -37,7 +37,7 @@ class EmptyMultibandTile(val cols:Int, val rows:Int, val cellType:CellType, val 
 
   override def band(bandIndex: Int): Tile = EmptyMultibandTile.empty(cellType, cols, rows)
 
-  override def bands: Vector[Tile] = Vector[Tile]()
+  override def bands: Vector[Tile] = List.fill(bandCount)(EmptyMultibandTile.empty(cellType, cols, rows)).toVector
 
   override def subsetBands(bandSequence: Seq[Int]): MultibandTile = new EmptyMultibandTile(cols,rows,cellType,bandSequence.size)
 
--- a/openeo-geotrellis/src/test/scala/org/openeo/geotrellis/EmptyMultibandTileTest.scala
+++ b/openeo-geotrellis/src/test/scala/org/openeo/geotrellis/EmptyMultibandTileTest.scala
@@ -25,23 +25,28 @@ object EmptyMultibandTileTest {
 }
 
 @RunWith(classOf[Parameterized])
-class EmptyMultibandTileTest(ct:CellType) {
+class EmptyMultibandTileTest(ct: CellType) {
 
   //@Parameterized.Parameter(value = 0) var ct: CellType = IntConstantNoDataCellType
-
 
 
   @Test
   def testCreateEmpty(): Unit = {
     val tile = EmptyMultibandTile.empty(ct, 10, 10)
-    assertEquals(NODATA,tile.get(0,0))
+    assertEquals(NODATA, tile.get(0, 0))
   }
 
   @Test
   def testCreate(): Unit = {
     val emptyMultibandTile: MultibandTile = new EmptyMultibandTile(10, 10, ct, 3)
     assertEquals(emptyMultibandTile.bandCount, 3)
-    assertEquals(NODATA, emptyMultibandTile.bands.head.get(0, 0))
-    assertEquals(NODATA, emptyMultibandTile.band(1).get(0, 0))
+    assertEquals(emptyMultibandTile.bands.size, 3)
+    for (band <- emptyMultibandTile.bands) {
+      // Test the corners:
+      assertEquals(NODATA, band.get(0, 0))
+      assertEquals(NODATA, band.get(0, band.rows - 1))
+      assertEquals(NODATA, band.get(band.cols - 1, 0))
+      assertEquals(NODATA, band.get(band.cols - 1, band.rows - 1))
+    }
   }
 }
--- a/openeo-geotrellis/src/test/scala/org/openeo/geotrellis/EmptyMultibandTileTest.scala
+++ b/openeo-geotrellis/src/test/scala/org/openeo/geotrellis/EmptyMultibandTileTest.scala
@@ -1,9 +1,8 @@
 package org.openeo.geotrellis
 
 import java.util
-
 import geotrellis.raster.CellType.constantNoDataCellTypes
-import geotrellis.raster.{CellType, FloatUserDefinedNoDataCellType, IntUserDefinedNoDataCellType, NODATA, UByteUserDefinedNoDataCellType, UShortUserDefinedNoDataCellType}
+import geotrellis.raster.{CellType, FloatUserDefinedNoDataCellType, IntUserDefinedNoDataCellType, MultibandTile, NODATA, Tile, UByteUserDefinedNoDataCellType, UShortUserDefinedNoDataCellType}
 import org.junit.Assert.assertEquals
 import org.junit.Test
 import org.junit.runner.RunWith
@@ -38,4 +37,11 @@ class EmptyMultibandTileTest(ct:CellType) {
     assertEquals(NODATA,tile.get(0,0))
   }
 
+  @Test
+  def testCreate(): Unit = {
+    val emptyMultibandTile: MultibandTile = new EmptyMultibandTile(10, 10, ct, 3)
+    assertEquals(emptyMultibandTile.bandCount, 3)
+    assertEquals(NODATA, emptyMultibandTile.bands.head.get(0, 0))
+    assertEquals(NODATA, emptyMultibandTile.band(1).get(0, 0))
+  }
 }
--- a/openeo-geotrellis/src/test/scala/org/openeo/geotrellis/LayerFixtures.scala
+++ b/openeo-geotrellis/src/test/scala/org/openeo/geotrellis/LayerFixtures.scala
@@ -113,7 +113,7 @@ object LayerFixtures {
    * Returns an RDD with tiles that switch between data and noData.
    * patternScale 2 gives [0 0 T T 0 0 T T] (where 0 is noData, and T is a data tile)
    */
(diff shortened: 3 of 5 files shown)


EmileSonneveld:
@bossie Jenkins gives errors like: Sentinel3PyramidFactoryTest.compareReferenceImage: `/data/MTDA/TERRASCOPE_Sentinel3/COG/SY_2_VG1/S3A/2020/07/01/S3A_SY_2_VG1____20200701T120000_20200702T120000_20200707T122712_EUROPE____________LN2_O_NT_002/S3A_SY_2_VG1____20200701T120000_20200702T120000_20200707T122712_EUROPE____________LN2_O_NT_002.SEN3/S3A_SY_2_VG1____20200701T120000_20200702T120000_20200707T122712_EUROPE____________LN2_O_NT_002.SEN3/MIR.tif (No such file or directory)`

But this test passes locally. Do you think it will work when merged on `develop`?

bossie:
The majority of the jenkinsslaves have that mount but some don't so it's a matter of luck. I'll check with s-ops but don't let this keep you from merging into develop.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/EmptyMultibandTile.scala
```diff
@@ -37,7 +37,7 @@ class EmptyMultibandTile(val cols:Int, val rows:Int, val cellType:CellType, val
 
   override def band(bandIndex: Int): Tile = EmptyMultibandTile.empty(cellType, cols, rows)
 
-  override def bands: Vector[Tile] = Vector[Tile]()
+  override def bands: Vector[Tile] = List.fill(bandCount)(EmptyMultibandTile.empty(cellType, cols, rows)).toVector
```
bossie: Just use `Vector.fill` iso/ `List.fill` + `toVector`.

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/EmptyMultibandTileTest.scala
```diff
@@ -38,4 +37,11 @@ class EmptyMultibandTileTest(ct:CellType) {
     assertEquals(NODATA,tile.get(0,0))
   }
 
+  @Test
+  def testCreate(): Unit = {
+    val emptyMultibandTile: MultibandTile = new EmptyMultibandTile(10, 10, ct, 3)
+    assertEquals(emptyMultibandTile.bandCount, 3)
+    assertEquals(NODATA, emptyMultibandTile.bands.head.get(0, 0))
+    assertEquals(NODATA, emptyMultibandTile.band(1).get(0, 0))
```
bossie: Assert that `emptyMultibandTile.bands.size` also == 3 and test for NODATA in a loop.

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/LayerFixtures.scala
```diff
@@ -67,7 +68,81 @@ object LayerFixtures {
...
+                                            tileLayout: TileLayout,
+                                            cellType: CellType = IntConstantNoDataCellType,
+                                            extent: Extent = defaultCRS.worldExtent,
+                                          )(implicit sc: SparkContext): MultibandTileLayerRDD[SpaceTimeKey] = {
+
+    val layout = LayoutDefinition(extent, tileLayout)
+    val keyBounds = {
+      val GridBounds(colMin, rowMin, colMax, rowMax) = layout.mapTransform(extent)
+      val minTime = tiles.minBy(_._2)._2
+      val maxTime = tiles.maxBy(_._2)._2
```
bossie: Leverage pattern matching e.g. `val (_, minTime) = tiles.minBy { case (_, timestamp) => timestamp }`

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/LayerFixtures.scala
```diff
@@ -67,7 +68,81 @@ object LayerFixtures {
...
+      extent = extent,
+      cols = tileLayout.layoutCols,
+      rows = tileLayout.layoutRows
+    )
+
+    val tileBounds = re.gridBoundsFor(extent)
+
+    val tilesArr = tiles.toArray
+    val tmsTiles = tileBounds.coordsIter.zip(tilesArr.iterator).map {
+      case ((col, row), (tile, time)) => (SpaceTimeKey(col, row, time), tile)
```
bossie: Not sure what this means: it's zipping something with only a spatial dimension with something that has a time dimension as well. I believe this implicitly assumes that there is only a single timestamp.

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/LayerFixtures.scala
```diff
@@ -67,7 +68,81 @@ object LayerFixtures {
...
+    }
+
+    new ContextRDD(sc.parallelize(tmsTiles.toSeq), metadata)
+  }
+
+  /**
+   * Practical for testing nodata tiles.
+   */
+  def buildSpatioTemporalDataCubePattern(tilingFactor: Int = 1, patternScale: Int = 1): ContextRDD[SpaceTimeKey, MultibandTile, TileLayerMetadata[SpaceTimeKey]] = {
```
bossie: Can probably be written shorter as: `MultibandTileLayerRDD[SpaceTimeKey]`.

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/LayerFixtures.scala
bossie: Also: no need to materialize `tiles` to an `Array`: `tiles` has a `toIterator` method.

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/MergeCubesSpec.scala
```diff
@@ -19,13 +21,14 @@ import java.time.format.DateTimeFormatter
 import java.util
 import scala.collection.JavaConverters._
 import scala.collection.mutable.ListBuffer
+import scala.language.implicitConversions
```
bossie: Can be removed.

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/MergeCubesSpec.scala
```diff
@@ -121,6 +144,63 @@ class MergeCubesSpec {
...
   }
 
+
+  @ParameterizedTest
+  @MethodSource(Array("testMergeCubesTiledNodataArguments"))
+  def testMergeCubesTiledNodata(aggregateR: AggregationType.Value,
+                                aggregateG: AggregationType.Value,
+                                aggregateB: AggregationType.Value,
+                               ): Unit = {
+    val path = "tmp/testMergeCubesTiledNodata/" + aggregateR + aggregateG + aggregateB + "/"
```
bossie: JUnit 5 has `@TempDir` for this.

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/MergeCubesSpec.scala
```diff
@@ -121,6 +144,63 @@ class MergeCubesSpec {
...
+          intervals,
+          labels,
+          TestOpenEOProcessScriptBuilder.createMedian(true),
+          java.util.Collections.emptyMap()
+        )
+        // composite = p.filterEmptyTile(composite)
+        val tmp2 = new ContextRDD(composite, composite.metadata)
+        tmp2
+      }
+    }
```
bossie: I would have opted for an OO approach of defining a base trait with a single method (~ `agg`) and three implementing classes/objects to replace these enum values that don't have behavior.

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/MergeCubesSpec.scala
```diff
@@ -121,6 +144,63 @@ class MergeCubesSpec {
     assertTrue(mse < 0.1)
   }
 
+
+  @ParameterizedTest
+  @MethodSource(Array("testMergeCubesTiledNodataArguments"))
+  def testMergeCubesTiledNodata(aggregateR: AggregationType.Value,
```
bossie: Consider adding some comments about how the test works.

bossie (changes requested):
Added remarks inline.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/EmptyMultibandTile.scala
EmileSonneveld: nice. fixed

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/LayerFixtures.scala
EmileSonneveld: changed

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/MergeCubesSpec.scala
EmileSonneveld: indeed. Intellij did not highlight it for me.

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/MergeCubesSpec.scala
EmileSonneveld: I like to use the ./tmp folder, as it is easaly findable in the Jenkins workspace, and it it gitignored

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/MergeCubesSpec.scala
EmileSonneveld: done

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/EmptyMultibandTileTest.scala
EmileSonneveld: done

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/LayerFixtures.scala
EmileSonneveld: Nice, that works

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/LayerFixtures.scala
EmileSonneveld: True! I made it explicit that the date is the same across the RDD.

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/MergeCubesSpec.scala
EmileSonneveld: Sounds like more code, no? Ok if I leave it like this?

PR #268, opened 2024-02-27
Netcdf multiple samples a day

### LINKED ISSUE #257: geotiff + netcdf: support writing sub-day files
I propose to detect if timestamps have hours and or minutes, and in that case use a more accurate timestamp or resolution.

Here:
https://github.com/Open-EO/openeo-geotrellis-extensions/blob/d618910335cb0e5e7d6e00d5903091d410802015/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/geotiff/package.scala#L127

and here:

https://github.com/Open-EO/openeo-geotrellis-extensions/blob/01b67c85f27ff59fde22872a53a13306b79eb6f9/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/netcdf/NetCDFRDDWriter.scala#L130

EmileSonneveld:
For tiff and netcdf.
When saving the tiles, we usually round the date to daily resolution.
Now, when there is a date that does not exactly match midnight, we assume the temporal resolution is higher, and we save the results with seconds in resolution.

For tiff export, the ':' in the date part is replaced with '-', to avoid problems on Windows file systems: Example `openEO_2017-01-01T01-00-00Z.tif`

I used `ISO_ZONED_DATE_TIME` in the filename. Following an example I found in `geotiff/package.scale`


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/geotiff/package.scala
```diff
@@ -682,14 +700,28 @@ package object geotiff {
     val layout = rdd.metadata.layout
     val crs = rdd.metadata.crs
 
+    val temporalResolution = if (rdd.keys.filter({
+      case key: SpaceTimeKey =>
+        // true if not exactly n days:
+        Duration.between(fixedTimeOffset, key.time).getSeconds % secondsPerDay != 0
+      case _ =>
+        false
+    }).isEmpty()) TemporalResolution.days else TemporalResolution.seconds
```
jdries: I'm afraid that doing this on the rdd triggers an extra job in Spark.
Suggestion would be to just choose the resolution when you create the filename. Consequence is that in theory, filename format could change.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/geotiff/package.scala
```diff
@@ -95,6 +99,14 @@ package object geotiff {
     val croppedExtent: Extent = preProcessResult._2
     val preprocessedRdd: RDD[(SpaceTimeKey, MultibandTile)] with Metadata[TileLayerMetadata[SpaceTimeKey]] = preProcessResult._3
 
+    val temporalResolution = if (preprocessedRdd.keys.filter({
+      case key: SpaceTimeKey =>
+        // true if not exactly n days:
+        Duration.between(fixedTimeOffset, key.time).getSeconds % secondsPerDay != 0
+      case _ =>
+        false
+    }).isEmpty()) TemporalResolution.days else TemporalResolution.seconds
```
jdries: I'm afraid that doing this on the rdd triggers an extra job in Spark.
Suggestion would be to just choose the resolution when you create the filename. Consequence is that in theory, filename format could change.

jdries (changes requested):
2 comments in geotiff, would it be possible to already apply changes of netcdfrddwriter to develop branch?

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/geotiff/package.scala
EmileSonneveld: Merged the NetCDF part, the Tiff part is  split to a different MR: https://github.com/Open-EO/openeo-geotrellis-extensions/pull/269

PR #272, opened 2024-03-04
663 aggregate spatial can not be used as input to vector to raster

JeroenVerstraelen:


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/vector/VectorCubeMethods.scala
```diff
@@ -41,56 +57,128 @@ object VectorCubeMethods {
...
+  }
+
+  /**
+   * Convert a vector datacube to a raster datacube with a time dimension.
+   *
+   * @param path: Path to the geojson file.
+   * @param t
+   * @return A raster datacube.
+   */
+  def vectorToRasterTemporalGeneric(path: String, targetLayout: LayoutDefinition, targetCRS: CRS, spatialBounds: KeyBounds[SpatialKey]): MultibandTileLayerRDD[SpaceTimeKey] = {
```
jdries: the 'generic' in the name probably was meant to say that it works for bot spatial and spacetime.
Given that this does not seem to be possible, I would drop the 'generic'.

jdries:
one small comment, but other than that I would merge this quickly so that we can get this through the builds

PR #283, opened 2024-04-17
add ids to raster_to_vector's geojson output Open-EO/openeo-geopyspar…

JeroenVerstraelen:
…k-driver#578

Add id's to the features returned by raster_to_vector in the following format:
`band0_{yyyyMMdd}_{feature_index}`
For example:
`band0_20240130_5`

`band0` can later be expanded to include the actual band names. Currently raster_to_vector only vectorizes the first band of the input cube.


jdries (approved):
approved, isn't there some existing unit test that can be extended to check on the presence of these generated id's?

PR #296, opened 2024-05-28
support "in" operator

### LINKED ISSUE #776: make property filter operators explicit
The way the `properties` argument of `load_collection` is currently passed throughout the application is problematic. Historically it assumes an exact match of a single value ("eq") but that no longer holds; the actual operator is lost along the way (in most cases).

Some related issues:

- Blocks https://github.com/Open-EO/openeo-python-driver/issues/145:
> the actual comparison operator is stripped somewhere in the pipeline

- A TODO for https://github.com/Open-EO/openeo-geotrellis-extensions/issues/39 re: what seems like a way to support cloud cover "lte" (not sure how OSCARS interprets this):
https://github.com/Open-EO/openeo-geopyspark-driver/blob/f2c07399ae0482640d1c1d9d5cfdfc001b7b1d81/openeogeotrellis/layercatalog.py#L306-L307

- Technical debt from https://github.com/Open-EO/openeo-opensearch-client/issues/32 where "eq" is actually interpreted as "in"; this issue has more in-depth knowledge re: this technical debt.

bossie:
Seemed like a good way to test drive the "in" operator before supporting it in Terrascope S2L2A but no time to address all of https://github.com/Open-EO/openeo-geopyspark-driver/issues/776.

### DIFF
--- a/geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/Criteria.scala
+++ b/geotrellis-sentinelhub/src/main/scala/org/openeo/geotrellissentinelhub/Criteria.scala
@@ -5,7 +5,7 @@ import scala.collection.JavaConverters._
 
 object Criteria {
   private def isPropagated(metadataProperty: String): Boolean =
-    !Set("provider:backend", "federation:backends").contains(metadataProperty)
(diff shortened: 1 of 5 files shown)


bossie:
A side-effect of this PR is that `org.openeo.geotrellissentinelhub.PyramidFactoryTest` now uses JUnit 5.

PR #302, opened 2024-06-04
Use more robust withRetryAfterRetries. 

### LINKED ISSUE #299: load_stac to stac.openeo.vito.be: 429 Too Many Requests
The following snippet triggers a 429 error. A backoff strategy or polling mechanism could avoid that
```python
import openeo

url = "https://openeo.dataspace.copernicus.eu"
connection = openeo.connect(url).authenticate_oidc()

EXTENT = dict(zip(["west", "south", "east", "north"],
                  [5.318868004541495, 50.628576059801816, 5.3334400271343725, 50.637843899562576]))
EXTENT['crs'] = "EPSG:4326"
STARTDATE = '2022-01-01'
ENDDATE = '2022-03-31'

meteo_cube = connection.load_stac("https://stac.openeo.vito.be/collections/agera5_daily", spatial_extent=EXTENT,
                                  temporal_extent=[STARTDATE, ENDDATE], bands=["2m_temperature_mean"])
job = meteo_cube.create_job()
job.start_and_wait()
```
```
py4j.protocol.Py4JJavaError: An error occurred while calling z:org.openeo.geotrellis.geotiff.package.saveRDDTemporal. : org.apache.spark.SparkException: Job aborted due to stage failure: Task 18 in stage 4.0 failed 4 times, most recent failure: Lost task 18.3 in stage 4.0 (TID 119) (10.42.110.101 executor 2): scalaj.http.HttpStatusException: 429 Error: HTTP/1.0 429 Too Many Requests
	at scalaj.http.HttpResponse.throwIf(Http.scala:156)
	at scalaj.http.HttpResponse.throwError(Http.scala:168)
	at org.openeo.geotrellis.CustomizableHttpRangeReader.totalLength$lzycompute(CustomizableHttpRangeReader.scala:22)
	at org.openeo.geotrellis.CustomizableHttpRangeReader.totalLength(CustomizableHttpRangeReader.scala:10)
	at geotrellis.util.StreamingByteReader.ensureChunk(StreamingByteReader.scala:109)
	at geotrellis.util.StreamingByteReader.get(StreamingByteReader.scala:130)
	at geotrellis.raster.io.geotiff.reader.GeoTiffInfo$.read(GeoTiffInfo.scala:127)
	at geotrellis.raster.io.geotiff.reader.GeoTiffReader$.readMultiband(GeoTiffReader.scala:211)
	at geotrellis.raster.geotiff.GeoTiffReprojectRasterSource.$anonfun$tiff$1(GeoTiffReprojectRasterSource.scala:46)
	at scala.Option.getOrElse(Option.scala:189)
	at geotrellis.raster.geotiff.GeoTiffReprojectRasterSource.tiff$lzycompute(GeoTiffReprojectRasterSource.scala:43)
	at geotrellis.raster.geotiff.GeoTiffReprojectRasterSource.tiff(GeoTiffReprojectRasterSource.scala:40)

```

EmileSonneveld:


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -56,7 +58,7 @@ package object geotrellis {
         .maxBackoffTime(Duration.ofMillis(15))
         .build()
     val retryPolicy =
-      RetryPolicy.defaultRetryPolicy()
+      software.amazon.awssdk.core.retry.RetryPolicy.defaultRetryPolicy()
```
bossie: At the top:

```scala
import software.amazon.awssdk.core.retry.{RetryPolicy => AwsRetryPolicy}
import net.jodah.failsafe.{RetryPolicy => FailsafeRetryPolicy}
```

then:

```suggestion
      AwsRetryPolicy.defaultRetryPolicy()
```

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -158,28 +160,67 @@ package object geotrellis {
...
-      }
-      timeUnit.sleep(amount)
-      attempt(retries - 1, (amount * 2, timeUnit)) {
-        action
-      }
+
+  /**
+   * taken from DefaultProcessApi
+   */
+  def withRetryAfterRetries[R](context: String)(fn: => HttpResponse[R])(implicit logger: Logger): HttpResponse[R] = {
```
bossie:
```suggestion
  def withRetryAfterRetries[R](context: String)(httpResponse: => HttpResponse[R])(implicit logger: Logger): HttpResponse[R] = {
```

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -158,28 +160,67 @@ package object geotrellis {
...
+    var lastResponse: Option[HttpResponse[R]] = None
+
+    val retryable: Throwable => Boolean = {
+      case HttpStatusException(400, _, responseBody) if responseBody.contains("Request body should be non-empty.") => true
```
bossie: This `"Request body should be non-empty."` response is SHub-specific so this line can be removed.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -158,28 +160,67 @@ package object geotrellis {
...
+      case HttpStatusException(statusCode, _, responseBody) if statusCode >= 500
+        && !responseBody.contains("newLimit > capacity") && !responseBody.contains("Illegal request to https") => true
```
bossie: Also SHub-specific, just retry on every 500 response:

```suggestion
      case HttpStatusException(statusCode, _, responseBody) if statusCode >= 500 => true
```

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -158,28 +160,67 @@ package object geotrellis {
...
+      case _: SocketTimeoutException => true
+      case _: CirceException => true
```
bossie: Handling `CirceException` is only applicable if the response is being parsed by Circe so can be removed.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -158,28 +160,67 @@ package object geotrellis {
...
-      timeUnit.sleep(amount)
-      attempt(retries - 1, (amount * 2, timeUnit)) {
-        action
-      }
+
+  /**
+   * taken from DefaultProcessApi
+   */
+  def withRetryAfterRetries[R](context: String)(fn: => HttpResponse[R])(implicit logger: Logger): HttpResponse[R] = {
+    var lastResponse: Option[HttpResponse[R]] = None
```
bossie: Unnecessary, your last response is an argument of `withDelay` (see further on).

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -158,28 +160,67 @@ package object geotrellis {
...
+      case HttpStatusException(statusCode, _, responseBody) if statusCode >= 500
+        && !responseBody.contains("newLimit > capacity") && !responseBody.contains("Illegal request to https") => true
+      case _: SocketTimeoutException => true
+      case _: CirceException => true
+      case e => logger.error(s"Not attempting to retry unrecoverable error in context: $context", e); false
     }
 
-    attempt(retries = 4, delay = (30, SECONDS)) {
-      action
+    val shakyConnectionRetryPolicy = new net.jodah.failsafe.RetryPolicy[HttpResponse[R]]()
```
bossie:
```suggestion
    val shakyConnectionRetryPolicy = new FailsafeRetryPolicy[HttpResponse[R]]()
```

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -158,28 +160,67 @@ package object geotrellis {
...
+        val e = execution.getFailure
+        logger.error(s"Failed after ${execution.getAttemptCount} attempt(s) in context: $context", e)
+      })
+
+    val isRateLimitingResponse: Throwable => Boolean = {
+      case HttpStatusException(429, _, _) => true
+      case _ => false
     }
+
+    var retryAfterSecondsCounter = 20.0
```
bossie: Unnecessary to keep state yourself, delay with back-off can be derived from `executionContext` argument of `withDelay` (see further on).

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -158,28 +160,67 @@ package object geotrellis {
...
+    val rateLimitingRetryPolicy = new net.jodah.failsafe.RetryPolicy[HttpResponse[R]]()
```
bossie:
```suggestion
    val rateLimitingRetryPolicy = new FailsafeRetryPolicy[HttpResponse[R]]()
```

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -158,28 +160,67 @@ package object geotrellis {
...
+      .handleIf(isRateLimitingResponse.asJava)
+      .withMaxAttempts(5)
+      .withDelay((_: HttpResponse[R], _: HttpStatusException, _: ExecutionContext) => {
```
bossie:
```suggestion
      .withDelay((lastResponse: HttpResponse[R], _: HttpStatusException, executionContext: ExecutionContext) => {
```

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -158,28 +160,67 @@ package object geotrellis {
...
+        val retryAfterSeconds = lastResponse match {
+          case None => retryAfterSecondsCounter.toLong
+          case Some(x) => x
+            .headers
+            .find { case (header, _) => header equalsIgnoreCase "retry-after" }
+            .map { case (_, values) => values.head.toLong }
+            .getOrElse(retryAfterSecondsCounter.toLong)
+        }
+        retryAfterSecondsCounter *= 1.6
+        Duration.ofSeconds(retryAfterSeconds)
```
bossie: Try the `retry-after` header first and if it's not present, calculate the delay from `executionContext.getAttemptCount`.

Will probably involve something like a combination of your initial delay of 20s and `math.pow(1.6, executionContext.getAttemptCount - 1)` or something (untested).

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -158,28 +160,67 @@ package object geotrellis {
...
+      })
+      .onRetryScheduled((retry: ExecutionScheduledEvent[scalaj.http.HttpResponse[_root_.geotrellis.raster.io.geotiff.MultibandGeoTiff]]) => {
```
bossie:
```suggestion
      .onRetryScheduled((retry: ExecutionScheduledEvent[HttpResponse[R]]) => {
```

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -158,28 +160,67 @@ package object geotrellis {
...
+        logger.warn(s"Scheduled retry within ${retry.getDelay} because of 429 response in context: $context")
+      })
+
+    Failsafe
+      .`with`(java.util.Arrays.asList(shakyConnectionRetryPolicy, rateLimitingRetryPolicy))
+      .get(() => {
+        val res = fn
+        lastResponse = Some(res)
+        res.throwError
```
bossie:
```suggestion
.get(() => httpResponse.throwError)
```

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/CustomizableHttpRangeReader.scala
```diff
@@ -1,17 +1,27 @@
...
 import geotrellis.util.RangeReader
+import org.slf4j.{Logger, LoggerFactory}
 import scalaj.http.HttpRequest
 
 import scala.util.Try
 
+
 // copied from geotrellis.util.HttpRangeReader as that one is not extensible
 class CustomizableHttpRangeReader(request: HttpRequest, useHeadRequest: Boolean) extends RangeReader {
+  private implicit val logger: Logger = LoggerFactory.getLogger(this.getClass)
```
bossie: Unnecessary to create a new `Logger` instance every time. Create a `CustomizableHttpRangeReader` companion object with a `logger` field:

```suggestion
  private implicit val logger: Logger = LoggerFactory.getLogger(classOf[CustomizableHttpRangeReader])
```

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/CustomizableHttpRangeReader.scala
```diff
@@ -1,17 +1,27 @@
...
+
+
+
   override lazy val totalLength: Long = {
     val headers = if (useHeadRequest) {
-      withRetries { request.method("HEAD").asString }
+      withRetryAfterRetries("totalLength") {
```
bossie:
```suggestion
      withRetryAfterRetries("HEAD totalLength") {
```

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/CustomizableHttpRangeReader.scala
```diff
@@ -1,17 +1,27 @@
...
+        request.method("HEAD").asString
+      }
     } else {
-      withRetries { request.method("GET").execute { _ => "" } }
+      withRetryAfterRetries("totalLength") {
```
bossie:
```suggestion
      withRetryAfterRetries("GET totalLength") {
```

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/CustomizableHttpRangeReader.scala
```diff
@@ -37,21 +47,13 @@ class CustomizableHttpRangeReader(request: HttpRequest, useHeadRequest: Boolean)
...
-     * Otherwise, the server SHOULD return a response with a status of 206
-     * (Partial Content) containing the satisfiable ranges of the entity-body."
-     * https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html
-     */
-    require(res.code != 416,
-      s"Server unable to generate the byte range between ${start} and ${start + length} for ${request.url}")
+    res.throwError
 
     require(res.is2xx,
       s"While reading ${request.url}, server returned status code ${res.code} with type ${res.contentType} and body ${new String(res.body)}")
```
bossie: I guess this can be removed then.

bossie (changes requested):
Remarks inline.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
EmileSonneveld: changed

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/CustomizableHttpRangeReader.scala
EmileSonneveld: I think `withRetryAfterRetries` could still give non 2xx responses, like 300. So I left this just to be sure

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
EmileSonneveld: Renamed to `httpResponseCallback`. Is is still a callback

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
EmileSonneveld: Refactored to not be dependant on `throwError`

EmileSonneveld:
The timeouts where already `connTimeoutMs = 10000, readTimeoutMs = 40000` when downloading the tiff. What is standard on multiple locations in OpenEO

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
bossie: I don't think this approach is better as you now have duplicate code and added complexity.

If the policy of `withRetryAfterRetries` is: the `httpResponseCallback` that you (the client programmer) pass in will only be retried on exceptions, then you have to make sure that his `httpResponseCallback` raises those exceptions = what `.throwError` does.

I would do away with the `handleResultIf` and just throw exceptions from within the `httpResponseCallback` (see elsewhere).

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -158,28 +158,65 @@ package object geotrellis {
...
-      timeUnit.sleep(amount)
-      attempt(retries - 1, (amount * 2, timeUnit)) {
-        action
-      }
+
+  /**
+   * taken from DefaultProcessApi
+   */
+  def withRetryAfterRetries[R](context: String)(httpResponseCallback: => HttpResponse[R])(implicit logger: Logger): HttpResponse[R] = {
+    val retryableResult: HttpResponse[R] => Boolean = r => r.is5xx
```
bossie, line 167: Duplicate code.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -158,28 +158,65 @@ package object geotrellis {
...
+          "result code: " + attempt.getLastResult.code
+        } else ""
+        logger.warn(s"Attempt ${attempt.getAttemptCount} failed in context: '$context' " + msg)
+      })
+      .onFailure((execution: ExecutionCompletedEvent[HttpResponse[R]]) => {
+        val e = execution.getFailure
+        logger.error(s"Failed after ${execution.getAttemptCount} attempt(s) in context: '$context'" + e.getMessage)
+      })
+
+    val isRateLimitingResponse: HttpResponse[R] => Boolean = r => r.code == 429 // similar to isRateLimitingException
```
bossie: Duplicate code.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/CustomizableHttpRangeReader.scala
```diff
@@ -1,17 +1,28 @@
...
+
 // copied from geotrellis.util.HttpRangeReader as that one is not extensible
 class CustomizableHttpRangeReader(request: HttpRequest, useHeadRequest: Boolean) extends RangeReader {
+  import CustomizableHttpRangeReader._
+
   override lazy val totalLength: Long = {
     val headers = if (useHeadRequest) {
-      withRetries { request.method("HEAD").asString }
+      withRetryAfterRetries("HEAD totalLength") {
+        request.method("HEAD").asString
```
bossie, line 20: Move `headers.throwError` (below) to this callback and retry only on exceptions.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/CustomizableHttpRangeReader.scala
```diff
@@ -1,17 +1,28 @@
...
+      }
     } else {
-      withRetries { request.method("GET").execute { _ => "" } }
+      withRetryAfterRetries("GET totalLength") {
```
bossie: Even better: put a single `withRetryAfterRetries("totalLength")` around this `if` instead of two separate ones.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/CustomizableHttpRangeReader.scala
bossie: If you make sure that you follow redirects (this is already so because of `CustomizableHttpRangeReaderProvider`) then there are no 3xx responses and you just have to consider success and error responses.

In that case you can move the `.throwError` to the callback and remove the `is2xx` check entirely.

bossie (changes requested):
I have the feeling that just moving `.throwError` to the callbacks and only handle exceptions in `withRetryAfterRetries` will simplify things.

Detailed remarks inline.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
EmileSonneveld: With the `throwError` flow it is hacky to access the `Retry-After` header. Ok to only keep the without throwError flow?

bossie: Quite right, you don't have the response headers in that case. :thinking:

EmileSonneveld: There was no real need to support the .throwError logic. Support dropped and duplicate code cleaned up!

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/CustomizableHttpRangeReader.scala
EmileSonneveld: Refactored to not be dependant on `throwError`

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/CustomizableHttpRangeReader.scala
EmileSonneveld: I tested redirects and the 2xx check is indeed not needed. Removed

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -158,28 +158,63 @@ package object geotrellis {
...
+        val e = execution.getFailure
+        logger.error(s"Failed after ${execution.getAttemptCount} attempt(s) in context: '$context'" + e.getMessage)
+      })
+
+    val isRateLimitingResponse: HttpResponse[R] => Boolean = r => r.code == 429
+    val isRateLimitingException: Throwable => Boolean = _ => false // handle errors in shakyConnectionRetryPolicy
+    val rateLimitingRetryPolicy = new FailsafeRetryPolicy[HttpResponse[R]]()
+      .handleResultIf(isRateLimitingResponse.asJava)
+      .handleIf(isRateLimitingException.asJava)
+      .withMaxAttempts(5)
```
bossie:
```suggestion
    val isRateLimitingResponse: (HttpResponse[R], Throwable) => Boolean =
      (response, _ /* ignore exceptions, those are handled in shakyConnectionRetryPolicy */) => response.code == 429

    val rateLimitingRetryPolicy = new FailsafeRetryPolicy[HttpResponse[R]]()
      .handleIf(isRateLimitingResponse.asJava)
```

bossie: Untested but I think this more clearly conveys that you're suppressing the default behavior of retrying on exceptions.

bossie (changes requested):
Clarity issue, otherwise fine.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
EmileSonneveld: That works indeed. Got trough in normal case, and fails clearly when .throwException is used in callback

PR #310, opened 2024-06-28
Dateline

### LINKED ISSUE #279: no results for bbox near date line
This process graph will return an empty GeoTiff for sync requests and no assets for batch jobs:

```json
{
  "process_graph": {
    "loadcollection1": {
      "process_id": "load_collection",
      "arguments": {
        "bands": [
          "B02",
          "B03",
          "B04",
          "B08"
        ],
        "id": "SENTINEL2_L2A",
        "spatial_extent": {
          "west": 178.7384,
          "east": 178.8548,
          "south": 70.769,
          "north": 70.8254
        },
        "temporal_extent": [
          "2024-04-02",
          "2024-04-03"
        ]
      }
    },
    "saveresult1": {
      "process_id": "save_result",
      "arguments": {
        "data": {
          "from_node": "loadcollection1"
        },
        "format": "GTiff",
        "options": {}
      },
      "result": true
    }
  }
}
```

Internal ref: CDSE-96

EmileSonneveld:
The issue was an intersection calculation losing a lot of precision due to reprojecting and back.
Now the intersection between the feature and the requested extent is calculated in the target extent crs.

This made that some tests now load a smaller piece of the feature, on some places `{FloatConstantNoDataArrayTile@16026} ArrayTile(256,256,float32) `became `{PaddedTile@15829} PaddedTile(ArrayTile(140,122,float32),17,134,256,256)`


jdries (changes requested):
Can you also add a test where the feature extent is outside of the valid bounds of target CRS? For instance, a feature with global extent, and target extent being some utm zone.
I think this is the reason for going via EPSG:4326 to compute the intersection safely.

EmileSonneveld:
Added a test for feature in LatLng and target extent in UTM.
Had to change the code too. This comment describes the change:

> We convert both extents to a common CRS before taking the intersection.
> If one of the CRSes can cover the whole world (non-UTM), this will be used as common CRS.
> We give priority to use the target CRS as common one, because the intersection will be converted to it anyway

EmileSonneveld:
There where already  many cases where 2 different UTM extentds where used. Confirmed by logging

jdries (changes requested):
it now assumes that utm is the only projection that has this issue.
EPSG:3035 is another example that we work with a lot.
It could be that the valid area of a crs is in fact available, maybe we can check on that.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
```diff
@@ -1415,13 +1416,27 @@ class FileLayerProvider private(openSearch: OpenSearchClient, openSearchCollecti
...
-      val targetExtentInLatLon = targetExtent.reproject(feature.crs.get)
-      val featureExtentInLatLon = feature.rasterExtent.get.reproject(feature.crs.get,LatLng)
-
-      val intersection = featureExtentInLatLon.intersection(targetExtentInLatLon).map(_.buffer(1.0)).getOrElse(featureExtentInLatLon)
-      val tmp = expandToCellSize(intersection.reproject(LatLng, targetExtent.crs), theResolution)
+      val featureIsUTM = feature.crs.get.proj4jCrs.getProjection.getName == "utm"
+      val targetIsUTM = targetExtent.crs.proj4jCrs.getProjection.getName == "utm"
+      val commonCrs = if (!targetIsUTM) targetExtent.crs
+      else if (!featureIsUTM) feature.crs.get
+      else targetExtent.crs // Avoid conversion imprecision by intersecting directly in the target CRS
```
jdries, line 1434: Assuming a method 'isValidExtent(crs,extent:ProjectedExtent)' exists to check if a certain ProjectedExtent is valid in a certain crs exists, pseudo code below.

The idea is to keep on using LatLng as common crs in case of uncertainty, making sure that we stick to the current behaviour as default option. The isValidExtent method can be implemented with support for LatLng, UTM, EPSG:3035, WebMercator to cover some common cases, until the proj4 scala API allows a proper lookup of the valid extent for a given projection.

For UTM, make sure to test for cases with neighbouring utm zones, as this is very common. 

```suggestion
      val commonCrs = if (isValidExtent(targetExtent.crs,featureExtent)) targetExtent.crs
      else if (isValidExtent(feature.crs,targetExtent)) feature.crs.get
      else LatLng
```

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
```diff
@@ -1415,13 +1416,27 @@ class FileLayerProvider private(openSearch: OpenSearchClient, openSearchCollecti
...
+
+      val featureExtentInCommonCRS = feature.rasterExtent.get.reproject(feature.crs.get, commonCrs)
+      val targetExtentInCommonCRS = targetExtent.extent.reproject(targetExtent.crs, commonCrs)
+
+      val intersection = featureExtentInCommonCRS.intersection(targetExtentInCommonCRS).map(_.buffer(1.0))
+      val intersectionTargetCrs = intersection match {
+        case None => targetExtent.extent.reproject(targetExtent.crs, targetExtent.crs)
```
jdries: I'm not sure if we should keep this case, can you open a new issue to look into simply discarding the feature when there is no intersection?
Adding a log line could help to reveal if this actually occurs in production, or just an edge case in the code.

EmileSonneveld: Just checked, it gets in this case when an extent gets wrapped badly around the antimeridian after reproject. 
We would be more in control if we double check the project result and wrap it manually if needed
image

EmileSonneveld: Maybe only changing this could be the fix.
The UTM projected to latlon becomes extent is `[-179, 177]`, which is wrong.
It should be represented as:
`[177, 181]` or maybe even `[-183, -179]`
But I guess some code prefers to keep everything within `[-180, 180]`, this would need some extra checking.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
EmileSonneveld: Ah, I see. That should be safe indeed. The current implementation assumes that in a request UTM will always be valid when projected to something else, and that a random CRS can not safely be projected to UTM.

Side note: I found that UTM has a huge tolerance for extends that are horizontally outside of their bounds. There were successful jobs where the extent was multiple UTM zones from the correct one and still worked. So an UTM check for validity mostly looks at the vertical coordinates.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
```diff
@@ -1410,18 +1411,54 @@ class FileLayerProvider private(openSearch: OpenSearchClient, openSearchCollecti
...
+      val intersectionTargetCrs = intersection match {
+        case None =>
+          logger.warn(s"Feature extent $featureExtentInCommonCRS and target extent $targetExtentInCommonCRS do not intersect.")
+          // TODO: Discard the feature?
+          targetExtent.extent
+        case Some(value) => value.reproject(commonCrs, targetExtent.crs)
+      }
+      val tmp = expandToCellSize(intersectionTargetCrs, theResolution)
+      if (!healthCheckExtent(ProjectedExtent(tmp, targetExtent.crs))) {
+        throw new IllegalArgumentException(s"Feature extent $featureExtentInCommonCRS is invalid in target CRS ${targetExtent.crs}.")
```
jdries, line 1460: This exception message can be more informative: 
* add process id (load_collection/load_stac)
* add feature id
* use STAC terminology, so 'Item' or 'Asset' instead of 'Feature',

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
```diff
@@ -219,6 +222,55 @@ package object geotrellis {
...
+    // positive width and height is already enforced in Extent.
+    val horizontal_tolerance = 1.1
+    val polygonIsUTM = projectedExtent.crs.proj4jCrs.getProjection.getName == "utm"
+    if (polygonIsUTM) {
+      // This is an extent that has the highest sensible values for northern and/or southern hemisphere UTM zones
+      val utmProjectedBoundsOriginal = Extent(166021.44, 0000000.00, 833978.56, 10000000)
+      val utmProjectedBounds = utmProjectedBoundsOriginal.buffer(
+        utmProjectedBoundsOriginal.width * horizontal_tolerance, 0)
+      if (!projectedExtent.extent.intersects(utmProjectedBounds)) {
+        logger.warn("healthCheckExtent dangerous extent: " + projectedExtent)
```
jdries, line 238: Make this logging more informative towards openEO user. Same as above.
If the purpose of the health check is to avoid using a 'dangerous' extent, than we may also want to avoid alarming the user.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
jdries: The final version is still assuming that a target crs that is not UTM is always safe to use?
Please add a test where you use a local reference system like EPSG:3812 and a 'feature' with global bounds.

jdries (commented):
Some important comments still, please check!

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
EmileSonneveld: The latest version does use the isValidExtent trick:
https://github.com/Open-EO/openeo-geotrellis-extensions/blob/9c2ed933a60d6c217675d0c4a333734b93c77270/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala#L1441-L1443


readDataCubeWithOpensearchClientUTM has a big LatLon feature (-180.0014880952381, -59.99851190473005, 179.998511906399, 80.00148809523809) where an UTM region is requested. I just added readDataCubeWithOpensearchClientLAEA for good measure.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
EmileSonneveld: The latlon warp is avoided by using `safeReproject()`

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
EmileSonneveld: Reworked to 
`healthCheckExtentAssert(featureExtentInCommonCRS, s"Item extent (${feature.id}) should be valid in common CRS: ")` and `healthCheckExtentAssert(targetExtent, s"Target extent should be valid: ")`

The later will print something like this when it found an issue:
`Target extent should be valid: Extent not within its CRS limits: ProjectedExtent(Extent(649630.0, 1.212245E7, 684180.0, 1.219141E7),EPSG:32631)`

[New PR](https://github.com/Open-EO/openeo-geotrellis-extensions/pull/399/files#diff-891944eb54f3fdfd4e94fc238e95eb5b3817d73b21bc7efa4d95ce3efa44dae2R1429)

For normal usage, I did not see the asserts get triggered tough. They are there just in case. I'll monitor the logs

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
EmileSonneveld: Also the feature will now get discarded when not intersecting.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
EmileSonneveld: I did not find a quick way to log the process id too.  -> added a comment to add this in the future: https://github.com/Open-EO/openeo-geotrellis-extensions/pull/399/commits/24c81c560066b065632bd2542823c09e1f644ea1

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/package.scala
EmileSonneveld: Changed logs to a more user friendly:
`Target extent should be valid: Extent not within its CRS limits: ProjectedExtent(Extent(649630.0, 1.212245E7, 684180.0, 1.219141E7),EPSG:32631)`

Issue #196, opened 2023-08-21
half-open intervals for shub based collections


jdries:

bossie:
Related: https://github.com/Open-EO/openeo-geopyspark-driver/issues/34

bossie:
As discussed: to avoid having to change a lot of examples and such, it was decided to, for the moment, deviate from the OpenEO spec and allow start_time == end_time instead of returning an error.

PR #314, opened 2024-08-13
196 batch job errors on vlcc prod

bossie:
Various fixes for [#196](https://github.com/eu-cdse/openeo-cdse-infra/issues/196).


bossie:
- exponential backoff for efficiency
- removed outer retry, more attempts for inner retry
- support for soft errors (but still no reliable way to trigger them)

PR #315, opened 2024-08-21
One tiff per band

### LINKED ISSUE #309: job-option to save results as one asset per band
For large projects, where the output is expected to be one file per s2 tile, having all the bands in one file would create too large files.
Thus the need for a job option to save the results in different files per band, as is often the case for other large products like the copernicus HRL VPP products.
Different assets should still be included in one stac item if possible. 

Implementation tasks:

- [ ] new geotiff writing approach in scala
- [ ] check STAC metadata, also the items!

## Scala implementation

I think the best place to start from is the code that now creates a tiff per date:
https://github.com/Open-EO/openeo-geotrellis-extensions/blob/83e9b63e699b6196d60c96257d7212cd89b972dd/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/geotiff/package.scala#L95

In the first 'map' where we compress tiles, the bandIndex will always be 0, but we will have to add the band name to the key, which is used to group the tuples in the rdd later on.

By including the band name in the rdd key, we will get compressed bytes per date per band, and can simply write those into a geotiff in a way that is almost identical to the current approach, except for generating the right filename of course.

EmileSonneveld:


jdries (approved):
approved, please merge like this to avoid further delay.
Instead of the deepclone, in scala sometimes people add a 'copy' method, that allows to create a copy with certain class members overridden.
Will be interesting to see how this method works on larger files.

EmileSonneveld:
Ok, I'll add it as TODO in the code and merge

PR #322, opened 2024-09-19
apply_dimension(dim=bands) => do not filter out all nodata tiles

### LINKED ISSUE #141: Investigate NaN pixels in DEM
By accident stumbled upon a weird DEM result for a patch extraction in Russia, generated in job `j-240830089c57467a8873ee79c0c5a6d2`. `COP-DEM` is a non-temporal band but gets replicated along the `t` axis. However, some nodata patches appear, **but only in the first temporal element**. Visually in RGB the first 2 temporal bands look like:

Image

And a timeseries in one of these red patches looks like:

Image

We should find out why this first element is nodata.

jdries:
do not filter out all nodata tiles, but replace them with 'empty' tile. This avoids issues in subsequent merge_cubes with spatial cube or linear interpolation over time.

https://github.com/WorldCereal/worldcereal-classification/issues/141


kvantricht:
@jdries what would be the timeline to get this in production?

jdries:
@kvantricht  unit tests seem to have succeeded, so:
- if urgent, I can merge immediately
- if not yet super urgent, it would be better if I manage to cover this with a unit test as well, and then merge

kvantricht:
@jdries if this is in production in let's say a week from now, it should be fine for us to still test some things before the webinar 17 October. Probably qualifies as not too urgent then?

PR #335, opened 2024-10-16
Move file to results

### LINKED ISSUE #329: separate_asset_per_band gives some empty tiffs
Example graph that uses `separate_asset_per_band` and has empty tiff files: `j-241009a45a764383a3a3db1453b9881f`
Making the batchjob write to S3 directly instead of the fuse mount avoids this issue. 
Need to check if fsync also avoids the issue: https://github.com/yandex-cloud/geesefs/blob/master/README.md?plain=1#L279-L299

https://teams.microsoft.com/l/message/19:2941a270bf8e48a2a8e8a23975051c11@thread.skype/1728034200593?tenantId=9e2777ed-8237-4ab9-9278-2c144d6f6da3&groupId=8c9c739d-2544-4def-8cd4-b65970551b70&parentMessageId=1728034200593&teamName=Unit%20TAP&channelName=openEO-users&createdTime=1728034200593

EmileSonneveld:
There is already a version that only tries the move twice on staging. This will make 5 tries and is a bit more thread-safe

### DIFF
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/geotiff/package.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/geotiff/package.scala
@@ -700,7 +700,6 @@ package object geotiff {
                                    croppedExtent: Option[Extent], cropDimensions: Option[java.util.ArrayList[Int]],
                                    compression: Compression, formatOptions: Option[GTiffOptions] = None
                                   ) = {
-    this.logger.info("stitchAndWriteToTiff tiles.size: " + tiles.size) // Remove before release
     val raster: Raster[MultibandTile] = ContextSeq(tiles, layout).stitch()
 
     val re = raster.rasterExtent
@@ -743,9 +742,7 @@ package object geotiff {
     ) {
       geotiff = geotiff.withOverviews(NearestNeighbor, List(4, 8, 16))
     }
-    val res = writeGeoTiff(geotiff, filePath)
-    this.logger.info("stitchAndWriteToTiff writeGeoTiff done. filePath: " + filePath) // Remove before release
-    res
+    writeGeoTiff(geotiff, filePath)
   }
 
   def saveSamples(rdd: MultibandTileLayerRDD[SpaceTimeKey],
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/geotiff/package.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/geotiff/package.scala
@@ -39,7 +39,7 @@ import java.time.format.DateTimeFormatter
 import java.util.{ArrayList, Collections, Map, List => JList}
 import scala.collection.JavaConverters._
 import scala.reflect._
-import scala.util.control.Breaks.break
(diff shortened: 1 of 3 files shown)


EmileSonneveld:
This should also fix the error in integration tests.
`openeo.rest.OpenEoApiError: [500] Internal: Server error: PermissionError(13, 'Permission denied') (ref: r-241016413b7f4eae8ed54d00388a426f)`

`Files.createTempFile` makes an empty file with by default very tight permissions. After a move the permissions stay the same. Now, no default empty file is created, and GDAL writes a file with more sensible permissions

EmileSonneveld:
@bossie  had a nice idea to avoid the retry logic:
Let executors write to unique paths, and send it to the driver. The driver chooses what files get moved to the final location, and then removes files from failed executors.

bossie:
It's just that doing it twice apparently does not suffice, and why would doing it 5 times fix it?

If the problem is that executors are getting in each other's way (be it because of speculation or otherwise) maybe we should just avoid that.

I thought maybe this could work:
 - driver passes the save directory to its executors;
 - executors each write to their file(s) in said directory but append some unique suffix to the filename so you get e.g. openEO_20241017Z_abc123.tif and return the file names to the driver;
 - the driver only gets one value back (in this case: per date), regardless of how many executors attempted to write it;
 - the driver strips off the winning suffix, moves the file and deletes the losing files.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/geotiff/package.scala
```diff
@@ -838,31 +836,47 @@ package object geotiff {
...
 
   }
 
+  def moveOverwriteWithRetries(oldPath: Path, newPath: Path): Unit = {
+    var try_count = 1
+    breakable {
+      while (true) {
+        try {
+          if (newPath.toFile.exists()) {
+            // It might be a partial result of a previous failing task.
```
jdries, line 864: Isn't it safer to just assume that the file is correctly written, or being written when the path already exists?

EmileSonneveld: I did not encounter it, but feel that the file might be a partial written result before that executor went OOM. 
OOM is a frequent reason why a second executor is started

EmileSonneveld:
The unique filenames solution is probably more robust, but a file move on S3 is more heavy. It will need to download and upload the file again, giving some delays and maybe will take a lot of memory again. (Not more than during job execution I guess, but stil)

jdries:
> * the driver strips off the winning suffix, moves the file and deletes the losing files.

This also sounds feasible. Only have to be careful with the fact that users want to provide arbitrary asset names. Of course, if the suffix uses a fixed number of characters, then I guess it's fool proof?

EmileSonneveld:
I'll start implementing that one. 
If re-enabling the fusemount is urgent, we can maybe get this one trough already. Or at least the permissions fix

jdries:
@EmileSonneveld  yes perhaps merge this one already. I'd like to avoid too much delay.

PR #336, opened 2024-10-16
Filepath per band

EmileSonneveld:
Maybe best to not allow `filepath_per_band` on datacubes with temporal dimension to start.


EmileSonneveld:
Merged to be able to test in `openeo-geopyspark-driver`

PR #356, opened 2024-12-12
355 mergecubes uses default geotrellis partitioner in certain cases which causes oom errors

### LINKED ISSUE #355: mergecubes uses default geotrellis partitioner in certain cases which causes OOM errors
The default SpacePartitioner uses 16x16 tiles per partition which requires too much memory if there are many bands, dates or if the tiles are large. This commit introduces a partitioner that attempts to control the partition size (max records) based on a maximum memory footprint. 

The new partitioner in this special case can have 1,2,4,8,16,32,64,128,256,512, or 1024 records in a single partition based on the dimensions of a single MultiBandTile: tile size, number of bands, and celltype.

JeroenVerstraelen:
The default SpacePartitioner uses 16x16 tiles per partition which requires too much memory if there are many bands, dates or if the tiles are large. This commit introduces a partitioner that attempts to control the partition size (max records) based on a maximum memory footprint.

The new partitioner in this special case can have 1,2,4,8,16,32,64,128,256,512, or 1024 records in a single partition based on the dimensions of a single MultiBandTile: tile size, number of bands, and celltype.

### DIFF
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcesses.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcesses.scala
@@ -730,11 +730,16 @@ class OpenEOProcesses extends Serializable {
     } else {
       // At least one partitioner is undefined.
       logger.info(s"Merging cubes with partitioners: ${leftCube.partitioner} - ${rightCube.partitioner} - many band case detected: $manyBands")
-      val nrBands = leftCount.getOrElse(1) + rightCount.getOrElse(1)
-      val outputCellType = maybeCellType(leftCube).getOrElse(DoubleCellType).union(maybeCellType(rightCube).getOrElse(DoubleCellType))
-      val tileSize = maybeTileSize(leftCube).getOrElse(128*128)
-      val newIndex = getPartitionerIndexForMaxPartitionSize[K](nrBands, tileSize, outputCellType.bits)
-      SpacePartitioner[K](kb)(implicitly, implicitly, newIndex)
+      if(manyBands) {
+        val index: PartitionerIndex[K] = getManyBandsIndexGeneric[K]()
+        SpacePartitioner[K](kb)(implicitly,implicitly,index)
+      } else {
+        val nrBands = leftCount.getOrElse(1) + rightCount.getOrElse(1)
+        val outputCellType = maybeCellType(leftCube).getOrElse(DoubleCellType).union(maybeCellType(rightCube).getOrElse(DoubleCellType))
+        val tileSize = maybeTileSize(leftCube).getOrElse(128 * 128)
+        val newIndex = getPartitionerIndexForMaxPartitionSize[K](nrBands, tileSize, outputCellType.bits)
+        SpacePartitioner[K](kb)(implicitly, implicitly, newIndex)
+      }
     }
 
     val joinRdd =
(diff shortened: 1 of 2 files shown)


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcesses.scala
```diff
@@ -708,14 +719,13 @@ class OpenEOProcesses extends Serializable {
         SpacePartitioner[K](kb)
       }
     } else {
+      // At least one partitioner is undefined.
       logger.info(s"Merging cubes with partitioners: ${leftCube.partitioner} - ${rightCube.partitioner} - many band case detected: $manyBands")
-      if(manyBands) {
```
jdries, line 712: why the immediate removal of manyBands?
I would prefer a more incremental change that takes things one step at a time, allowing to release this one faster without too much worries.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcesses.scala
```diff
@@ -708,14 +719,13 @@ class OpenEOProcesses extends Serializable {
...
-        val index: PartitionerIndex[K] = getManyBandsIndexGeneric[K]()
-        SpacePartitioner[K](kb)(implicitly,implicitly,index)
-      }else{
-        SpacePartitioner[K](kb)
-      }
-
+      val nrBands = leftCount.getOrElse(1) + rightCount.getOrElse(1)
+      val outputCellType = maybeCellType(leftCube).getOrElse(DoubleCellType).union(maybeCellType(rightCube).getOrElse(DoubleCellType))
+      val tileSize = maybeTileSize(leftCube).getOrElse(128)
```
jdries: this is wrong: geotrellis tilesize is actually returning width*height, so something needs to change

jdries: checking out usage of tilesize in getPartitionerIndexForMaxPartitionSize, the default should perhaps be 128*128
some scaladoc might also help to make sure that interpretation is clear

jdries (changes requested):

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcesses.scala
JeroenVerstraelen: ManyBands returned either ByTileSpatialPartitioner or SpaceTimeByMonthPartitioner which are both special cases of the configurable partitioners so I thought it was safe. But of course the new partitioner has to be more thoroughly tested to ensure it doesn't actually exceed its maximum allowed memory size so I will revert the manyBands case.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcesses.scala
JeroenVerstraelen: Pushed a fix.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcesses.scala
JeroenVerstraelen: Pushed a fix.

jdries (approved):

Issue #352, opened 2024-11-28
store gdalinfo to separate asset for geotif


jdries:
For LCFM, certain quality checks rely on the gdalinfo output rather than the stac metadata. (Gdalinfo has  a few extra fields, like block size.)
The request is to dump that output to a file, and retain it together with the other assets.
If we do this dump on the executors, and upload also to S3 (if applicable), then it also solves the issue of avoiding a (more expensive) gdalinfo on the driver.
I would not add the gdalinfo dump as an asset (yet), just keep it on the storage. We can then consider to export it to workspace perhaps based on feature flag. (Trying to keep this an internal thing, avoiding that users really depend on it.)

EmileSonneveld:
Related; https://github.com/Open-EO/openeo-geopyspark-driver/issues/948

EmileSonneveld:
There are some differences in the `"wkt"` property, and under the` "proj:projjson"` property. For example:
`"$schema": "https://proj.org/schemas/v0.7/projjson.schema.json",`
`"$schema": "https://proj.org/schemas/v0.2/projjson.schema.json",`

There is only one gdalinfo in the container:
`docker run -it vito-docker.artifactory.vgt.vito.be/openeo-base:latest`
```
find / -name "gdalinfo"
/usr/share/bash-completion/completions/gdalinfo
/usr/bin/gdalinfo
```
Differences where negligible

EmileSonneveld:
Can't download the gdalinfo through the python client atm. I'll add it to the asset list
```python
filename = "openEO_2023-06-01Z.tif_gdalinfo.json"
url = f"/jobs/{job.job_id}/results/assets/{filename}"
response: requests.Response = connection.get(url)
with open(filename, "w") as file:
    file.write(response.text)
```

`openeo.rest.OpenEoApiError: [400] FilePathInvalid: File path is invalid: 'openEO_2023-06-01Z.tif_gdalinfo.json' not in ['openEO_2023-06-06Z.tif', 'openEO_2023-06-09Z.tif', 'openEO_2023-06-14Z.tif', 'openEO_2023-06-01Z.tif', 'openEO_2023-06-04Z.tif', 'openEO_2023-06-11Z.tif'] (ref: r-25011414360242cbbc3d2dbab35aad0a)
`

TODO: Check if they get transferred to export workspace after this.

EmileSonneveld:
Fixed workspace and download issue. It got deployed on staging already:
[...]

PR #358, opened 2024-12-17
Gdalinfo on executor

EmileSonneveld:
https://github.com/Open-EO/openeo-geotrellis-extensions/issues/352
Add `_gdalinfo.json` object next to tiff files.
Might needs something equivalent for NetCDF files.


EmileSonneveld:
I would merge when the develop branch has no errors. Otherwise this might add extra confusion

Issue #253, opened 2024-01-16
raster_to_vector creates artifacts due to tile regridding


JeroenVerstraelen:
Before converting the raster to vectors, the raster_to_vector method regrids the rdd to become one single tile (or multiple very large ones) if the rdd is too big:
https://github.com/Open-EO/openeo-geotrellis-extensions/blob/master/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcesses.scala#L482

When raster_to_vector is used in the [parcel delineation notebook](https://github.com/openEOPlatform/parcel-delineation/blob/main/Parcel%20delineation.ipynb) it appears that this regridding causes artifacts to appear.

Input:
![raster_to_vector_input](https://github.com/Open-EO/openeo-geotrellis-extensions/assets/3028262/a6a99c7e-8ed3-470e-9d15-47c1bf344ec7)

Output:
![raster_to_vector_issue](https://github.com/Open-EO/openeo-geotrellis-extensions/assets/3028262/0ba8217a-8e4a-47aa-8860-236216f65f4a)

jdries:
I pushed a simple fix and improved the unit test.
Still I noticed that vectorization did not seem to include the full original raster for some reason.
Run the unit test modified in this issue to observe current behaviour!

jdries:
Proposed solution: add some overlap to the tiles, then vectorize, then perform a union that merges overlapping polygons

jdries:
Pointers:

The fix starts at this location, where a 'regrid' operation is performed, but does not yet use an overlap:
https://github.com/Open-EO/openeo-geotrellis-extensions/blob/4eab4ca92cff3cbd720d9322814cd8496e38aaf6/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcesses.scala#L598

There already exists a better regrid operation, that allows to specify an overlap:
https://github.com/Open-EO/openeo-geotrellis-extensions/blob/4eab4ca92cff3cbd720d9322814cd8496e38aaf6/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcesses.scala#L1014

I suggest to start with a limited overlap, of 8 pixels. (More overlap uses more memory.)

After vectorization, to merge overlapping geometries, the geotrellis 'VectorJoin' operation seems most promising.

jdries:
Extra things to consider:

- the udf will assign differents labels to parts of the parcel that occur in different chunks
[...]

PR #360, opened 2024-12-20
issue253 Use extend with overlap with retile in vectorize

ElienVandermaesenVITO:
#253


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcesses.scala
```diff
@@ -617,13 +617,19 @@ class OpenEOProcesses extends Serializable {
     val maxExtent = datacube.metadata.extent
     //naive approach: combine tiles and hope that we don't exceed the max size
     //if we exceed the max, vectorize will run on separate tiles, and we'll need to merge results
-    val newCols = Math.min(256*20,layout.cols)
-    val newRows = Math.min(256*20,layout.rows)
+    val newCols = Math.min(256,layout.cols)
```
jdries: Here I suggest to still work with a larger tile size by default, maybe 256*4, avoiding that we make an immediate change from very coarse tiling to small tiles.
It is always possible that certain edge cases were not detected in our testing, and by doing it like this we do a more gradual rollout.
It is however important to ensure that the unit test also continues to work on the more complex case. Making this tile size configurable would be one way to achieve that.

jdries (changes requested):

jdries:
@ElienVandermaesenVITO I'm considering to merge this one, but is the remaining issue described or logged somewhere?

Issue #359, opened 2024-12-20
Scala version of process graph visitor


jdries:
In Python, we have this class:

https://github.com/Open-EO/openeo-geopyspark-driver/blob/c7c24656ec83174466a1a9c038b59811b3a297bc/openeogeotrellis/processgraphvisiting.py#L11

It walks over a process graph, and invokes callbacks on a scala object.
We don't have an equivalent in scala, and as a consequence always have to write out callback invocation manually, which is very tedious.

Can we port this class to scala, preferably preserving the separation between an abstract base class and an implementation that invokes callbacks on OpenEOProcessScriptBuilder?

The hope is that we can then store process graphs as json in the scala repo, and easily set up test scenario's based on that json.
Maybe we can even define scenario's entirely in json...

PR #362, opened 2024-12-24
Issue359 scala version of process graph visitor

ElienVandermaesenVITO:
#359


jdries (approved):
Approved, can't immediately spot important things to change.
I would merge this one, and then we can try to apply it to some of the existing test cases.

PR #365, opened 2025-01-14
issue989 implement BatchJobProgressListener.onApplicationEnd to filter the stages to only log long-running stages

ElienVandermaesenVITO:
issue https://github.com/Open-EO/openeo-geopyspark-driver/issues/989#issue-2777087987


jdries (approved):

jdries:
Merging this one as it looks good.
One area of doubt is if we should still print something at the start or end of a stage, just for progress monitoring.
Other question will be if it's perhaps better to print a single multi-line log at the end, but not sure if that will play nice. To be evaluated on staging.

PR #368, opened 2025-01-28
issue 842 add option to read zarr files in FileLayerProvider.deriveRasterSources.raterSource

ElienVandermaesenVITO:
issue https://github.com/Open-EO/openeo-geotrellis-extensions/issues/611#issue-2477314733

### DIFF
--- a/openeo-geotrellis/src/test/scala/org/openeo/geotrellis/layers/FileLayerProviderTest.scala
+++ b/openeo-geotrellis/src/test/scala/org/openeo/geotrellis/layers/FileLayerProviderTest.scala
@@ -1,5 +1,6 @@
 package org.openeo.geotrellis.layers
 
+import akka.http.scaladsl.server.PathMatcher0
 import cats.data.NonEmptyList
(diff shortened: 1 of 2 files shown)


openeo-geotrellis/src/test/scala/org/openeo/geotrellis/layers/FileLayerProviderTest.scala
```diff
@@ -986,6 +990,65 @@ class FileLayerProviderTest extends RasterMatchers{
...
+      }
+    } finally {
+      tarInputStream.close()
+    }
+  }
+
+  @Test
+  def testReadKeysToRasterSources(@TempDir tempDir: Path):Unit = {
+    val Filename = "zarrExample.tar"
+    val url = new URL(s"https://artifactory.vgt.vito.be/artifactory/testdata-public/openeo/geotrellis_extrensions/$Filename")
```
jdries, line 1026: zarr should normally also work fine over http, could we try that rather than the download?
In the case of this test, we would mostly be reading metadata, so actual data download is very limited.

Issue #370, opened 2025-02-05
resample_cube_spatial: preserve sparse partitioner


jdries:
This code:
https://github.com/Open-EO/openeo-geotrellis-extensions/blob/5ad6537b4b89661ebb1cb9e2364f8cb21cd0c97a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcesses.scala#L899

Does not preserve sparse partitioners, resulting in a stage with a very large number of tasks, taking lots of time.

Job id  j-25020519344849d9be024b2c6fa5cc8e on staging reveals the issue. The important part is to do an aggregate_spatial with sparse sampling over a Sentinel-1 cube resampled to 20m and Sentinel-2 bands at 10m.

The source partitioner:
`aggregate_temporal results in 380 keys, using partitioner index: SparseSpaceTimePartitioner 360 true with bounds KeyBounds(SpaceTimeKey(0,0,1533081600000),SpaceTimeKey(17,17,1583020800000))`

The target partitioner in this case is:
`SparseSpaceTimePartitioner 380 true with bounds KeyBounds(SpaceTimeKey(0,0,1533081600000),SpaceTimeKey(71,68,1583020800000))`

PR #371, opened 2025-02-07
issue 370 add extra support SparceSpaceTimePartitioner

ElienVandermaesenVITO:
#370

### DIFF
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcesses.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcesses.scala
@@ -880,6 +880,27 @@ class OpenEOProcesses extends Serializable {
     ContextRDD(filtered,data.metadata.copy(bounds = newBounds))
   }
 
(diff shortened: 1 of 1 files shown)


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcesses.scala
```diff
@@ -880,6 +880,27 @@ class OpenEOProcesses extends Serializable {
     ContextRDD(filtered,data.metadata.copy(bounds = newBounds))
   }
 
+  def transfromSparceSpaceTimePartition(data: MultibandTileLayerRDD[SpaceTimeKey], target: MultibandTileLayerRDD[SpaceTimeKey]):Option[SparseSpaceTimePartitioner]= {
+    val index = data.partitioner.get.asInstanceOf[SpacePartitioner[SpaceTimeKey]].index.asInstanceOf[SparseSpaceTimePartitioner]
```
jdries: @ElienVandermaesenVITO  I think  index.asInstanceOf[SparseSpaceTimePartitioner] will throw an exception if index is not the right type, which is very much possible. Please add extra if statement here to first check for the right type.

PR #383, opened 2025-02-25
issue846 implement addTemporal in TileRasterLayer

ElienVandermaesenVITO:
issue 846 https://github.com/Open-EO/openeo-geopyspark-driver/issues/846#issue-2480662467

### DIFF
--- a/geopyspark-geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
+++ b/geopyspark-geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
@@ -130,7 +130,7 @@ class SpatialTiledRasterLayer(
     SpatialTiledRasterLayer(zoom, tileLayer)
   }
 
-  def addTemporal(temporalKey:TemporalKey): TiledRasterLayer[SpaceTimeKey] = {
+  def toTemporalLayer(temporalKey:TemporalKey): TiledRasterLayer[SpaceTimeKey] = {
     val bounds = rdd.metadata.bounds.get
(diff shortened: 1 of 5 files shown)


geopyspark-geotrellis/src/main/scala/geopyspark/geotrellis/TemporalTiledRasterLayer.scala
```diff
@@ -277,6 +277,14 @@ class TemporalTiledRasterLayer(
     TemporalTiledRasterLayer(zoom, tileLayer)
   }
 
+  def addTemporal(temporalKey:TemporalKey): TiledRasterLayer[SpaceTimeKey] = {
+    val bounds = rdd.metadata.bounds.get
```
jdries: I'm not sure if we need this? The cube already has a temporal dimension in this case, so why would we add one again?

geopyspark-geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
```diff
@@ -130,6 +130,14 @@ class SpatialTiledRasterLayer(
     SpatialTiledRasterLayer(zoom, tileLayer)
   }
 
+  def addTemporal(temporalKey:TemporalKey): TiledRasterLayer[SpaceTimeKey] = {
```
jdries: What about the name? 'add' sounds like we simply add something, while we're actually doing a conversion from spatial to temporal.
Does 'asTemporal' sound reasonable? Or are there similar method names that we can align to?

jdries (changes requested):

jdries:
Great, thanks!

PR #391, opened 2025-03-11
Make sure requests to hr-vpp-products buckets are sent to the correct endpoint

pvbouwel:
This change allows to use different regions than waw3-1 in getCreoS3Client because the endpoint should be constructed depending on the region.

The SWIFT endpoint is only used as fallback if no explicit region is specified (RegionOne is used)


JeroenVerstraelen (approved):
LGTM

PR #404, opened 2025-03-28
Issue387 zarr output

### LINKED ISSUE #387: Zarr output
Proof of concept for Zarr output. This format has many options, and is still evolving.

We should create a writer that maximally benefits from the fact that Zarr can be written in parallel, because the format is simply a directory of many small files.

Here's how to do that from java:
https://jzarr.readthedocs.io/en/latest/amazonS3.html#write-to-an-s3-bucket

My proposal is to define a Zarr with chunk sizes that match the RDD that is being written. Each timestep is also a separate chunk.

This means that when invoking save_result, we have to create the root location in object storage, perhaps this could even be done from python? 
Then the actual data writing will happen on executors: Geotrellis tiles are converted to Zarr arrays, compressed, and written to object storage.

After doing all that, it is possible that the metadata still requires an update. Potentially this can be done by collecting metadata from the executors in the driver. (Only the metadata, data is too potentially too large!)

Note that georeferencing metadata will not yet be present, that's a separate task.

Excluded stuff:
- proper stac metadata
- georeferencing
- advanced format options to define chunking

Checklist:

- [x] changelog entry -> mark as experimental
- [x] new output format -> mark as experimental
- [x] new test suite
- [ ] does it work on both terrascope (posix) and cdse (object storage)?
- [x] log 2 or 3 follow up issues, what are good next steps?
- [ ] What does performance in production look like? This format is supposed to be efficient.
- [ ] What about storage size?

ElienVandermaesenVITO:
#387

### DIFF
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/zarr/ZarrWriter.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/zarr/ZarrWriter.scala
@@ -19,7 +19,7 @@ object ZarrWriter {
 
   def saveZarr[K: SpatialComponent: Boundable : ClassTag](rdd:MultibandTileLayerRDD[K],path:String, nBands:Int):Unit= {
     val groupPath = path + "/" + getGroupName(path)
-    writeZGroup(groupPath)
+    writeFile(groupPath,FILENAME_DOT_ZGROUP,null)
     val metadata = rdd.metadata
     val cellType = metadata.cellType
     val (zarrType: DataType,fillValue:Option[Number]) = toZarrType(cellType)
@@ -46,13 +46,13 @@ object ZarrWriter {
         (dist.zipWithIndex.toMap, dist.length+:shapeBands, 1+:chunkBands,true)
       case _ => (Map[Long,Int](),shapeBands,chunkBands, false)
     }
-    writeZAttr(groupPath, new dataAttribute(metadata,nBands,hasTemp))
+    writeFile(groupPath,FILENAME_DOT_ZATTRS, new dataAttribute(metadata,nBands,hasTemp))
     val zarrHeader = new ZarrHeader(shape, chunk, zarrType.toString, byteOrder, fillValue.getOrElse(0), compressor, ".")
-    writeZArray(groupPath, zarrHeader)
+    writeFile(groupPath, FILENAME_DOT_ZARRAY, zarrHeader)
 
     writeVariables(path,"x",xValues.toArray)
     writeVariables(path,"y",yValues.toArray)
-    writeZGroup(path)
+    writeFile(path,FILENAME_DOT_ZGROUP,null)
 
 
     rdd.foreach{ case (k, multibandTileLayer) =>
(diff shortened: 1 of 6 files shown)


openeo-geotrellis/src/test/scala/org/openeo/geotrellis/zarr/ZarrWriterTest.scala
```diff
@@ -0,0 +1,107 @@
...
+    val layoutCols = 8
+    val layoutRows = 4
+
+    val intImage = LayerFixtures.createTextImage( layoutCols*256, layoutRows*256)
+    val imageTile = ByteArrayTile(intImage,layoutCols*256, layoutRows*256)
+
+    val tileLayerRDD = TileLayerRDDBuilders.createMultibandTileLayerRDD(ZarrWriterTest.sc,MultibandTile(imageTile),TileLayout(layoutCols,layoutRows,256,256),LatLng)
+
+    val filename = (tempDir / "out.zarr").toString()
+    zarr.ZarrWriter.saveZarr(tileLayerRDD,filename,1)
```
jdries: After writing, can we do some assertions on what was written?
Perhaps just using JZarr library to open the file and check that expected variables and dimensions are there?

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/zarr/ZarrWriter.scala
```diff
@@ -0,0 +1,185 @@
...
+import java.io.{OutputStream, OutputStreamWriter}
+import java.nio.ByteOrder
+import java.nio.file.Paths
+import scala.jdk.CollectionConverters.mapAsJavaMapConverter
+import scala.reflect.ClassTag
+
+object ZarrWriter {
+
+  def saveZarr[K: SpatialComponent: Boundable : ClassTag](rdd:MultibandTileLayerRDD[K],path:String, nBands:Int):Unit= {
+    val groupPath = path + "/" + getGroupName(path)
```
jdries, line 21: Use java Path to construct platform independent path name?

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/zarr/ZarrWriter.scala
```diff
@@ -0,0 +1,185 @@
...
+    val (shapeBands,chunkBands) = if (nBands > 1) {
+      (Array[Int](nBands,metadata.rows.toInt, metadata.cols.toInt),Array[Int](1,tileRows, tileCols))
+    } else{
+      (Array[Int](metadata.rows.toInt, metadata.cols.toInt),Array[Int](tileRows, tileCols))
+    }
+
+
+    val xValues = for (x <- 0 until metadata.cols.toInt) yield metadata.extent.xmin + x * metadata.cellwidth + metadata.cellwidth / 2.0
+    val yValues = for (y <- 0 until metadata.rows.toInt) yield metadata.extent.ymax - y * metadata.cellheight - metadata.cellheight / 2.0
+    val keys = rdd.keys.collect()
```
jdries, line 41: This will trigger Spark to evaluate the full RDD, which is also done in the 'foreach' below, so creating 2 separate spark jobs.
It would be more efficient if one job could both write the data and also return the metadata to be collected on the driver, if that is possible. 

The problem is probably how to map a timestamp to an index? One strategy is to check if the PartitionerIndex is perhaps holding the temporal keys. 
Another strategy is to simply map each day to an index, assuming that Zarr does not mind that certain dates are empty? This would however fail to work for cubes like Sentinel-3, that have sub-day timestamps...

jdries: Suggestion: do not consider this a blocker, but create proper follow-up tickets?
Only the strategy to fetch keys from partitioner index is perhaps easy, using existing code:
https://github.com/Open-EO/openeo-geotrellis-extensions/blob/9984ae329e391f0dc9d2c45d14c7afbb4df79775/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcesses.scala#L322

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/zarr/ZarrWriter.scala
```diff
@@ -0,0 +1,185 @@
...
+      case ct: UByteUserDefinedNoDataCellType => (DataType.u1,Some(ct.widenedNoData.asInt))
+      case ct: ShortUserDefinedNoDataCellType => (DataType.i2,Some(ct.noDataValue))
+      case ct: UShortUserDefinedNoDataCellType => (DataType.u2,Some(ct.widenedNoData.asInt.toShort))
+      case ct: IntUserDefinedNoDataCellType => (DataType.i4,Some(ct.noDataValue))
+      case ct: FloatUserDefinedNoDataCellType => (DataType.f4,Some(ct.noDataValue))
+      case ct: DoubleUserDefinedNoDataCellType => (DataType.f8,Some(ct.noDataValue))
+    }
+  }
+
+  private def writeZArray(path: String, zarrHeader: ZarrHeader): Unit = {
```
jdries: there seems to be some code duplication in these 3 subsequent writing methods?

jdries (changes requested):

jdries (approved):

PR #412, opened 2025-04-09
382 logging went missing when running tests

### LINKED ISSUE #382: Logging went missing when running tests
For example when running `testWriteMultibandTemporalRDDWithGapsSeparateAssetPerBand`, but probably is an issue with any test.
The logs show:
```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
SLF4J: Failed to load class "org.slf4j.impl.StaticMDCBinder".
SLF4J: Defaulting to no-operation MDCAdapter implementation.
SLF4J: See http://www.slf4j.org/codes.html#no_static_mdc_binder for further details.
```
And then many normal logs are not shown anymore.

Fixing the issue here might fix it also for `async_task`: https://github.com/Open-EO/openeo-geopyspark-driver/issues/1054#issuecomment-2668538798

cc @bossie

EmileSonneveld:
This MR errors on Jenkins with a checkout error. Re-making it on `382-logging-went-missing-when-running-tests-20250409`


bossie:
@EmileSonneveld not sure what the checkout error is about but why did you create a new branch instead of creating a new PR from the existing branch?

EmileSonneveld:
@bossie I don't see the checkout error anymore too. I updated the other poms too. And ran `mvn clean install` on the openeo-logging submodule.
Do think merging this would fix the missing scala logs on dev/staging?

bossie:
Hard to say really. `geotrellis-extensions` on prod is pretty recent and that one does have proper logging.

PR #415, opened 2025-04-10
Issue #392 added support for "array_contains" in the OpenEOProcessScr…

### LINKED ISSUE #392: array_contains is not supported in the OpenEOProcessScriptBuilder
cc @VictorVerhaert

dsamaey:


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcessScriptBuilder.scala
```diff
@@ -633,6 +632,37 @@ class OpenEOProcessScriptBuilder {
...
 
+  private def arrayContains(arguments:java.util.Map[String,Object]) : OpenEOProcess = {
+    val value = getProcessArg("value")
+    val data = getProcessArg("data")
+
+    val arrayContainsProcess = (context: Map[String, Any]) => (tiles: Seq[Tile]) => {
+      val value_input: Seq[Tile] = evaluateToTiles(value, context, tiles)
+      val data_input: Seq[Tile] = evaluateToTiles(data, context, tiles)
+      if (value_input.size != 1) {
+        throw new IllegalArgumentException("The value argument of the array_contains function should resolve to exactly one input.")
```
EmileSonneveld: Can add to message: `f"got {value_input.size} inputs instead"` (But nitpick)

EmileSonneveld (commented):
Is there also an openEO process graph JSON that tests this feature somewhere? Might be nice to add to the ticket

openeo-geotrellis/src/test/java/org/openeo/geotrellis/TestOpenEOProcessScriptBuilder.java
```diff
@@ -1232,10 +1231,33 @@ public void testArrayFindNoMatch() {
...
+        OpenEOProcessScriptBuilder builder = new OpenEOProcessScriptBuilder();
+        Map<String, Object> arguments = Collections.emptyMap();
+
+        builder.expressionStart("array_contains", arguments);
+        builder.argumentStart("data");
+        builder.argumentEnd();
+        builder.constantArgument("value",0);
+        builder.expressionEnd("array_contains",arguments);
+
+        Function1<Seq<Tile>, Seq<Tile>> transformation = builder.generateFunction();
```
jdries: Can we also put an assert on: builder.resultingDataType()

array_contains returns a boolean, so we expect to get BitCellType there.
This may require a change here:

https://github.com/Open-EO/openeo-geotrellis-extensions/blob/5233412dafbb499aa49b86a85a1400c961b9ee59/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcessScriptBuilder.scala#L1223

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcessScriptBuilder.scala
```diff
@@ -633,6 +633,37 @@ class OpenEOProcessScriptBuilder {
...
+      val valueInput: Seq[Tile] = evaluateToTiles(value, context, tiles)
+      val dataInput: Seq[Tile] = evaluateToTiles(data, context, tiles)
+      if (valueInput.size != 1) {
+        throw new IllegalArgumentException(f"The value argument of the array_contains function should resolve to exactly one input, got ${valueInput.size}.")
+      }
+      val theValue = valueInput.head
+      val tile = MultibandTile(dataInput)
+
+      val mutableResult:MutableArrayTile = ArrayTile.empty(BitCellType,tile.cols,tile.rows)
+      for (column <- Range(0, tile.cols)) {
```
jdries: similar methods use cfor for this triple loop:
https://github.com/Open-EO/openeo-geotrellis-extensions/blob/5233412dafbb499aa49b86a85a1400c961b9ee59/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcessScriptBuilder.scala#L226

Wondering if we should apply the same pattern, or if this is all equivalent?

jdries (changes requested):

openeo-geotrellis/src/test/java/org/openeo/geotrellis/TestOpenEOProcessScriptBuilder.java
```diff
@@ -1232,10 +1231,36 @@ public void testArrayFindNoMatch() {
...
 
 
-        Tile emptyResult = transformation.apply(JavaConversions.asScalaBuffer(Arrays.asList(tile0, tile1))).head();
+        Tile emptyResult = transformation.apply(JavaConverters.asScalaBuffer(Arrays.asList(tile0, tile1))).head();
         assertTrue(emptyResult.isNoDataTile());
     }
 
+    @DisplayName("Test array_contains process")
+    @Test
+    public void testArrayContains() {
```
jdries: is it maybe better to only retain the 'new style' test based on process graph loading?
I would like to try and reduce the volume of test code, for maintainability.

dsamaey: removed the obsolete test

jdries (approved):
Nice, thanks!

PR #423, opened 2025-04-16
Dateline 20250416

EmileSonneveld:
The issue is that the refine step changes it's resolution depending on the length of the line segment.
Giving weird refinements like this (product 2020_M08_60WWC_1_0):
image

The bottom side of this rectangle caused an artifact

https://github.com/Open-EO/openeo-geopyspark-driver/issues/1072

### DIFF
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/ProjectedPolygons.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/ProjectedPolygons.scala
@@ -259,14 +259,14 @@ object ProjectedPolygons {
    * Inspired on reprojectExtentAsPolygon from geotrellis. I could not find an equivalent for polygons in geotrellis self:
    * https://github.com/pomadchin/geotrellis/blob/b071b33/vector/src/main/scala/geotrellis/vector/reproject/Reproject.scala#L94
    */
-  def reprojectPolygonRefined(polygon: Polygon, transform: Transform, relError: Double): Polygon = {
+  def reprojectPolygonRefined(polygon: Polygon, transform: Transform, absError: Double): Polygon = {
     var interiorRings = List[LineString]()
     for (ringNr <- 0 until polygon.getNumInteriorRing) {
       val shell = polygon.getInteriorRingN(ringNr).asInstanceOf[LineString]
-      interiorRings = interiorRings :+ reprojectRingRefined(shell, transform, relError)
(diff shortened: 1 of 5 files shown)


EmileSonneveld:
Now that I see, the lines that cross the antimeridian are not refined extra because they wrap around the world and are considered very long, allowing for a larger relative error.

An issue with the absolute error is that is will keep on refining on the antimeridian:
image

EmileSonneveld:
Added testRefineToLatLngOverAntimeridian with simple triangle over antimeridian. Before and after picture:
image
image

EmileSonneveld:
@jdries I merged this to go through integration tests over night. The idea is to do the refine in 0-360 values of LatLng if the polygon crosses the antimeridian. A trick also used for calculating boolean operations on LatLng

Many polygon tests are just manually checked in Q-GIS, I re-compared them after this fix, and will automate them soon

EmileSonneveld:
Integration tests did not run on this overnight. Starting now...

EmileSonneveld:
A code snippet that I wrote and did not use. Keeping it ffr:
```scala
  /**
   * To calculate the distance between 2 'x' coordinates,
   * they need to be wrapped in the same way around the antimeridian.
   * See unit test for example usage.
   */
  def wrap_closer_by(toWrap: Double, toMatch: Double): Double = {
    if (Math.abs(toWrap - toMatch) < 300) {
      // Only do something when the distance is remarkably large.
      return toWrap
    }
    val toWrap2 = to_min180_180_range(toWrap)
    if (toMatch < -90) {
      return toWrap2 - 360
    }
    if (toMatch > 90) {
      return toWrap2 + 360
    }
    toWrap
  }


  @Test
  def testWrapCloserBy(): Unit = {
    val toMatch = 175
    assertEquals(wrap_closer_by(-179, toMatch), 181)
    assertEquals(wrap_closer_by(179, toMatch), 179)
    assertEquals(wrap_closer_by(10, toMatch), 10)

    val toMatchOver = 185
    assertEquals(wrap_closer_by(-179, toMatchOver), 181)
    assertEquals(wrap_closer_by(179, toMatchOver), 179)
    assertEquals(wrap_closer_by(10, toMatchOver), 10)

    val toMatchNegative = -185
    assertEquals(wrap_closer_by(-179, toMatchNegative), -179)
    assertEquals(wrap_closer_by(179, toMatchNegative), -181)
    assertEquals(wrap_closer_by(10, toMatchNegative), 10)
  }
```

PR #425, opened 2025-04-18
mask: use custom outerJoin to try and preserve partitioner

### LINKED ISSUE #424: mask looses detailed partitioner index
The default mask process (when not applied at load_collection time) will reset the partitioner to a 'regular' one, instead of attempting to preserve the partitioner and detailed info such as keys.

The 'outerJoin' method used by merge_cubes can be a better solution for this.
The only extra is that the keybounds of the masking cube could perhaps be used to 'pre-filter' the partitioner of the incoming cube? (Even though according to the spec, the mask cube should cover the entire input cube...)

jdries:
https://github.com/Open-EO/openeo-geotrellis-extensions/issues/424


jdries:
I just did a newer PR that also providers a spacetime-spatial join that may work for this:
https://github.com/Open-EO/openeo-geotrellis-extensions/pull/852

PR #427, opened 2025-04-22
Cleanup / enhance some dateline tests

EmileSonneveld:
This PR only changes test code. Should have no effect on code that runs when deployed.

For: https://github.com/Open-EO/openeo-geopyspark-driver/issues/1072


.pre-commit-config.yaml
```diff
@@ -0,0 +1,19 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+# Enable with the following commands:
+#   python3 -m pip install pre-commit
+#   pre-commit install
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
```
EmileSonneveld, line 7: @bossie Pre commit ok? It should only do some light formatting on added files (removing trailing whitespace and makes files end with a newline). 
As a test I let ran on all files and `mvn clean install -Dmaven.test.skip=true` still worked

Issue #386, opened 2025-02-26
netcdf add_offset and scale_factor metadata


jdries:
We already support setting offset and scale in geotiff, but not yet in netcdf.

This is where it needs to be added:
https://github.com/Open-EO/openeo-geotrellis-extensions/blob/3a094219ecc2324cc5afb10063b33b6bd1cb77b8/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/netcdf/NetCDFRDDWriter.scala#L694

For geotiff, it is possible to set these options via the `bands_metadata` format option, which is in fact hidden:

https://github.com/Open-EO/openeo-geopyspark-driver/blob/8f25d19bf56469f5e5d794c2a728fa12118d1d03/openeogeotrellis/geopysparkdatacube.py#L1856

jacobs-tim:
According to CF conventions, client apps who read add_offset and scale_factor will unpack the data into the _datatype_ of the scale_factor and add_offset. So make sure that both are float32 or float64

If valid_range, valid_min and/or valid_max are also used, then those take the raw digital values (ie without applying scale/offset that convert from the raw digital to the physical values).

jdries:
Extra: I propose to have uniform naming and behaviour for raster scale and offset, independent of the file format selected.
Therefore, I suggest to use STAC naming, which is 'raster:scale' and 'raster:offset'. The raster extension has the full definition:
https://github.com/stac-extensions/raster
Our own file format parameters should explain this, and refer to stac raster extension.

jacobs-tim:
@jdries your proposal for format-agnostic/STAC-based naming works for me

EmileSonneveld:
Ticket that implemented offset and scale for geotiff:
https://github.com/Open-EO/openeo-geotrellis-extensions/issues/317

PR #428, opened 2025-04-22
Issue386 netcdf add offset and scale bandsmetadata

ElienVandermaesenVITO:
#386


jdries (changes requested):
The netcdf attribute names should be add_offset and scale_factor, apologies, this was a bit lost in the issue comments.
We follow CF-conventions:
https://cfconventions.org/cf-conventions/cf-conventions.html#units-multiples

jdries (approved):

PR #433, opened 2025-04-30
Issue #58 merge of spacetime and spatial cubes based on outerjoin

### LINKED ISSUE #58: merge_cubes spatial + spacetime generates intersection rather than union

dsamaey:


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcesses.scala
```diff
@@ -976,34 +976,14 @@ class OpenEOProcesses extends Serializable {
...
-          if(l.bandCount==0) {
-            r
-          }else if(r.bandCount==0) {
-            l
-          }
-          throw new IllegalArgumentException("Merging cubes with an overlap resolver is only supported when band counts are the same. I got: %d and %d".format(l.bandCount, r.bandCount))
-        }else{
-          MultibandTile(l.bands.zip(r.bands).map(t => binaryOp.apply(if(swapOperands){Seq(t._2, t._1)} else Seq(t._1, t._2))))
-        }
+    val resampledSpaceTime = leftCube.cartesian(resampled).filter(a => a._1._1.spatialKey == a._2._1).map(a => (a._1._1, a._2._2))
```
jdries: I'm afraid cartesian has pretty large performance/memory impact?
One way to find keys (if available) is via: org.openeo.geotrellis.OpenEOProcesses#findPartitionerKeys
Then we at least have an efficient implementation for some cases.

For the less efficient case, perhaps create an RDD of keys only, and then perform a join on spatialkey, rather than doing cartesian on full rdd's?

jdries:
Need to see how the work in this PR: https://github.com/Open-EO/openeo-geotrellis-extensions/pull/852 has impact.

Issue #441, opened 2025-05-07
load_stac reading data from custom object storage location: profiles


jdries:
load_stac can return S3 url's from different regions, requiring different endpoints and credentials.

Currently, we have some hacky things in place, and fall back to the SWIFT_URL environment variable:
https://github.com/Open-EO/openeo-geotrellis-extensions/blob/bc720e4448b6491df9304dc663e0cb160b3592d2/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/MultiClientRangeReaderProvider.scala#L47

Instead, it would be good to propagate info from STAC collection, and set the right region accordingly.

jdries:
Suggestion is to mimic the 'workspace' approach that we have in Python. So we create a singleton that tracks workspace info, allowing to retrieve details based on a bucket name.

PR #443, opened 2025-05-07
add singleton to track bucket/workspace information in scala

jdries:
https://github.com/Open-EO/openeo-geotrellis-extensions/issues/441

### DIFF
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/MultiClientRangeReaderProvider.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/MultiClientRangeReaderProvider.scala
@@ -3,6 +3,7 @@ package org.openeo.geotrellis
 import geotrellis.store.s3.AmazonS3URI
 import geotrellis.store.s3.util.{S3RangeReader, S3RangeReaderProvider}
 import org.openeo.geotrellis.creo.CreoS3Utils
+import org.openeo.workspace.WorkspaceRepository
 import software.amazon.awssdk.regions.Region
 import software.amazon.awssdk.services.s3.S3Client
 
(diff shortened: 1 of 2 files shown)


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/MultiClientRangeReaderProvider.scala
```diff
@@ -28,8 +29,12 @@ class MultiClientRangeReaderProvider extends S3RangeReaderProvider {
     val isCloudFerro = s3Endpoint != null &&
       (s3Endpoint.toLowerCase.contains("cloudferro") || s3Endpoint.toLowerCase.endsWith(".dataspace.copernicus.eu"))
 
+    val maybeWorkspace = WorkspaceRepository.get().getWorkspaceByBucket(s3Uri.getBucket)
```
pvbouwel, line 32: I don't believe this logic is best placed here. CreoS3Utils has helpers like `getCreoS3Client` so if we solve the problem here in MultiClientRangeReader we'd have to solve it in other places as well.  Ideally there is one factory for creating S3 clients. There are however multiple problems with `getCreoS3Client` and it's consumption:

1. It assumes the region is known or defaults to 'RegionOne'
  a. a hard coded default does not make sense imho
  b. the placeholder value would cause the endpoint to be resolved via an environment variable `SWIFT_URL` but it would not change the region. If sigv4 checking is done stricly then it would lead to authorization failures.
2. From a consumption part it often called in the same file without specifying an argument

I like the idea of a Workspace Repository. Python could provision the workspaces info that is potentially encountered. 

So a ClientFactory method that just takes a bucket name (or S3URI from which it can extract the bucket name)  use the workspace repository to resolve bucket to S3 details (region + endpoint or region+profile) and potentially falls back to the legacy resolution would be usable in both places.

openeo-geotrellis/src/main/scala/org/openeo/workspace/WorkspaceRepository.scala
```diff
@@ -0,0 +1,40 @@
...
+
+case class WorkspaceConfig(bucketName: String, bucketRegion: String, profile: Option[String], bucketEndpoint: Option[String])
+
+class WorkspaceRepository {
+    /**
+     * A repository to retrieve information about workspaces.
+     */
+
+    private val workspaces = scala.collection.mutable.Map[String, WorkspaceConfig]()
+    private val workspacesByBucket = scala.collection.mutable.Map[String, WorkspaceConfig]()
```
pvbouwel, line 19: Is there a particular reason not to just have a `Map[String, String]`. A 1-to-1 relationship between workspace and bucketname is by doing 2 lookups the WorkspaceConfig won't be duplicated.

openeo-geotrellis/src/main/scala/org/openeo/workspace/WorkspaceRepository.scala
```diff
@@ -0,0 +1,40 @@
...
+
+    def registerBucketDetails( workspaceId:String, bucketName: String,
```
pvbouwel, line 21: What will call these register functions? I guess the singleton is per JVM so needs to be called on driver and all executors?

For K8s we can easily make sure a file is present in the execution environment. Is the same true for YARN? Otherwise the loading can be done on initialization of the singleton.

pvbouwel: Are sync jobs a thing on YARN because? Because for jobs I guess at spark-submit time extra files can be handed.

pvbouwel (changes requested):
I like the idea of the PR but implementation wise there are some unknowns and I would have some things different

pvbouwel:
This would no longer be needed when https://github.com/Open-EO/openeo-geotrellis-extensions/pull/756 is available.

PR #452, opened 2025-05-15
add GTiffOptions to all paths

ElienVandermaesenVITO:
Is usefull for https://github.com/Open-EO/openeo-geopyspark-driver/issues/1151


EmileSonneveld (approved):
looks good to me. method overloading if fine for py4j. (Default arguments not, but they are not used)

Issue #419, opened 2025-04-12
tile size of the COG overviews is not as expected


JeroenVerstraelen:
Reporter: Joris Coddé (Teams)

The tile size of the COG overviews is not as expected for one result (node) of our processes.
 
While cog_validate runs fine on them, the problem is that the tile size for the overviews is too low (128), whereas we specify 256 via the save_result_options:
 
```
    save_result_options = {
        "filename_prefix": filename,
        "separate_asset_per_band": True,
        "attach_gdalinfo_assets": True,
        "file_metadata": file_metadata,
        "tile_size": 256,
    }
```
 
This makes that the current files cannot be ingested in the viewer (Sinergise specifies 256 as a lower limit, likely for efficiency reasons).
 
Now this is only for the GAMMA0-NB-OBS result of our sentinel 1 feature process, likely because it does not get a scale+offset on the save_result node. The other bands are proper (tile size 256 everywhere, but they have scale+offset defined).
Note that the tile size for the data itself is proper (as shown by gdalinfo), the problem is only for the overviews. So the difference is revealed by tiffinfo, not by gdalinfo.
 
Example id: j-250411151107457e9ea8e386bbf48a7d
Pg: https://gist.github.com/JeroenVerstraelen/a260562e75d7f5010de3bae604e6787a
 
It would be great that OpenEO output could be ingested directly by Sinergise's BYOC.
 
Note: of lower priority than the outstanding issues of the nodata gap and antimeridian.

JorisCod:
Here's a (more) minimal process graph:
j-2504121818464540881d8075a75cf16c
https://gist.github.com/JorisCod/2a4f0c1912a1713cfc9a075b780039df

It can likely be reduced more.

Btw I tried setting scale & offset to 1/0, but this does not change the result (i.e. it's not a workaround).
Cfr. j-2504121828164fef817b41773ad3f91f

jdries:
Based on minimal graph, it happens when saving a spatial datacube as one single tiff, with single asset per band.

The tile size adjustment now happens here:
https://github.com/Open-EO/openeo-geopyspark-driver/blob/83c04af1768690f824a098d650a340e5c27ba4ff/openeogeotrellis/geopysparkdatacube.py#L2003

I had actually expected it here, where retiling can happen as part of other preprocessing operations:
[...]

PR #454, opened 2025-05-15
Issue419 tile size in overviews

ElienVandermaesenVITO:
#419


jdries (changes requested):
looks good, can we also add test coverage?

jdries (approved):

Issue #459, opened 2025-05-21
areaInSquareMeters: "Cannot create a polygon with unclosed exterior"


EmileSonneveld:
Found by Kristof here: `j-25052107390741839bb6bf989db511cb`

EmileSonneveld:
This is deployed on CDSE prod, and the job runs there fine

PR #460, opened 2025-05-21
Unclosed exterior

EmileSonneveld:
https://github.com/Open-EO/openeo-geotrellis-extensions/issues/459


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/ProjectedPolygons.scala
```diff
@@ -243,7 +243,7 @@ object ProjectedPolygons {
    * @return
    */
   private def areaInSquareMeters(geometry: Geometry, crs: CRS): Double = {
-    val bounds = geometry.extent
+    val bounds = geometry.extent.reproject(crs, LatLng)
     val targetCrs = CRS.fromString(s"+proj=aea +lat_0=0 +lon_0=0 +lat_1=${bounds.ymin} +lat_2=${bounds.ymax} +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs")
```
EmileSonneveld, line 247: @jdries Does this look ok? Between lat_1 and lat_2 the CRS should give the minimal amount of distortion, but we specified it in random values first, while it should be a degree.

jdries (approved):

PR #462, opened 2025-05-22
graceful degradation writeToDisk feature

ElienVandermaesenVITO:


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/netcdf/NetCDFRDDWriter.scala
```diff
@@ -417,8 +417,7 @@ object NetCDFRDDWriter {
           }
           case t: Throwable =>
             logger.error(s"save_result netCDF: Failed to write sample: $name error: ${t.getMessage}", t)
-            throw t
-
+            (outputAsPath.toString,extent)
```
jdries, line 420: you will want to preserve more of the behaviour that we use in IOException? (Or is this just for testing something?)

jdries (changes requested):

PR #463, opened 2025-05-23
426 load collection option to not filter out empty tiles

### LINKED ISSUE #426: load_collection: Option to not filter out empty tiles
load_collection will simply remove tiles that are fully nodata, which is fine for a lot of use cases.
Some use cases however want to for instance perform a 'count', also including nodata. The filtering behaviour is problematic in that case, because you can't count something that simply is not there.

Example:
```
        cube.reduce_temporal(
            lambda x: eop.count(
                x,
            )
        )
        .apply(lambda x: if_(is_nan(x), 0, x))
```

### Proposed fix

We already have 'EmptyMultibandTile', which is now quite broadly supported in many places.
I suggest to add this, initially under feature flag or job option, so we do not break everything if there's unwanted side effects.

It's also possible that there's a performance impact.

JeroenVerstraelen:
- new retainNoDataTiles datacube parameter
- adds EmptyMultibandTile if all bands have `isNoDataTile` tiles
- reuses old retainNoDataTiles property from FileLayerProvider that was not used anymore

### DIFF
--- a/geotrellis-common/src/main/scala/org/openeo/geotrelliscommon/DataCubeParameters.scala
+++ b/geotrellis-common/src/main/scala/org/openeo/geotrelliscommon/DataCubeParameters.scala
@@ -33,6 +33,12 @@ class DataCubeParameters extends Serializable {
    */
   var maxPartitionSize: Option[Int] = None
 
+  /**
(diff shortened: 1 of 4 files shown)


jdries (approved):

HansVRP:
what's the best way we can check this fix?

PR #469, opened 2025-06-02
interpret int '1' as '1.0'

jdries:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/1232


jdries:
No longer relevant, fixed with other approach!

PR #474, opened 2025-06-11
Introduce RasterSourceProvider

### LINKED ISSUE #471: introduce RasterSourceProvider
This method:
org.openeo.geotrellis.layers.FileLayerProvider#deriveRasterSources
now supports creating multiple types of raster sources by inspecting the incoming Feature/Item.

I propose an approach where we introduce a RasterSourceProvider, which can declare if it supports a given incoming Item.

Preconfigured collections should be able to also set a fixed RasterSourceProvider, from the config file. So if we require a very custom provider for e.g. Sentinel-2 products, we can do exactly that, but in a relatively clean way.

Note: for the initial implementation, make sure to keep the old code path, allowing an 'opt-in' approach for the cases where we want to configure the new one.

jdries:
Example of a complex RasterSourceProvider for Sentinel-2, containing many of the 'ugly' parts.

This would need to be integrated into FileLayerProvider, here:
https://github.com/Open-EO/openeo-geotrellis-extensions/blob/a45a5ea058f6a03bcba9bc24b4c1e5cbc05d423b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala#L1527-L1526

Intentionally keeping the old code path the default one, allowing to gradually enable this.

This test could be one to try and run, it does not depend on the actual data:
org.openeo.geotrellis.layers.FileLayerProviderTest#testCreoNonNativeProjection


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/DefaultItemRasterSourceProvider.scala
```diff
@@ -0,0 +1,290 @@
...
+import org.openeo.geotrellis.layers.FileLayerProvider.{convertNetcdfLinksToGDALFormat, vsis3ToS3}
+import org.openeo.geotrelliscommon.DataCubeParameters
+import org.openeo.opensearch.{OpenSearchClient, OpenSearchResponses}
+import org.openeo.opensearch.OpenSearchResponses.{Feature, Link}
+import org.slf4j.{Logger, LoggerFactory}
+
+import java.net.URI
+import java.nio.file.Paths
+
+case class DefaultItemRasterSourceProvider(openSearch: OpenSearchClient, openSearchLinkTitles: NonEmptyList[String], rootPath: String,
```
jdries, line 21: We probably want to work out the dependency on OpenSearchClient in the long run?
(Somewhat in doubt myself if we should aim to remove it in this PR.)

jdries (commented):
For the review, I would like to merge this one asap, as otherwise we will get conflicts given that quite a bit of code is moved.
Are there parts or blocks that we should focus on, or that may have important changes with respect to original behaviour?
Was the resulting jar tested in a real world job already?

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/DefaultItemRasterSourceProvider.scala
dsamaey: Added https://github.com/Open-EO/openeo-geotrellis-extensions/issues/507 to do this later

PR #483, opened 2025-07-02
linear_scale_range: fix thresholds for type casting

soxofaan:
ref: https://github.com/Open-EO/openeo-geopyspark-driver/issues/1275

also see https://github.com/Open-EO/openeo-geopyspark-driver/commit/8e1e82e8dc4cd894538c49b141165a35aabeecf7


bossie (approved):
Will indeed cast to byte on `outputMax` == 254 but not for [1000, 1012]. :+1:

PR #484, opened 2025-07-02
issue 1151 add overviews to temporal geotiff

ElienVandermaesenVITO:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/1151


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/geotiff/package.scala
```diff
@@ -257,6 +257,14 @@ package object geotiff {
...
             DateTimeFormatter.ISO_ZONED_DATE_TIME.format(key.time).replace(":", "").replace("-", "")
           }
+          val overviews = if(formatOptions.overviews.toUpperCase == "ALL" || (formatOptions.overviews.toUpperCase == "AUTO" && (gridBounds.width>1024 || gridBounds.height>1024 )) ) {
+            val decimationFactor = 4
+            val rasterTile = Raster(tile,croppedExtent)
+            val resampleMethod = getOverviewResampleMethod(formatOptions)
+            val resample = rasterTile.resample(RasterExtent(rasterTile.extent,tileLayout.tileCols/decimationFactor,tileLayout.tileRows/decimationFactor),resampleMethod)
+            val croppedBytes = raster.CroppedTile(resample.tile, raster.GridBounds(0, 0, tileLayout.tileCols/decimationFactor - 1, tileLayout.tileRows/decimationFactor - 1)).toBytes()
+            theCompressor.compress(croppedBytes,0)
+          } else Array[Byte]()
```
EmileSonneveld: In an other overviews snippet `Nil` is used. Would that work here too?

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/geotiff/package.scala
```diff
@@ -257,6 +257,14 @@ package object geotiff {
             // ':' is not valid in a Windows filename
             DateTimeFormatter.ISO_ZONED_DATE_TIME.format(key.time).replace(":", "").replace("-", "")
           }
+          val overviews = if(formatOptions.overviews.toUpperCase == "ALL" || (formatOptions.overviews.toUpperCase == "AUTO" && (gridBounds.width>1024 || gridBounds.height>1024 )) ) {
+            val decimationFactor = 4
```
EmileSonneveld: why 4?
Would `Pyramid.up` work here, like in the other pyramid calculation?  https://github.com/Open-EO/openeo-geotrellis-extensions/blob/aaf9914e8933d3e716a6332c98e3c7ffa0b0632d/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/geotiff/package.scala#L551

ElienVandermaesenVITO: The `pyramid.up` works on a rdd, and is not possible here

jdries (approved):
Approved!
I would however ask the LCFM team to do a before/after comparison on one of their workflows, so they can see the impact on memory/performance.
If it's good enough for them, than it should be fine for us. 
Note that I believe this change will immediately enable overviews for all users. (Because 'AUTO' is default value in format options?) We could also make this more strict and only generate them if set to 'ALL'.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/geotiff/package.scala
```diff
@@ -839,6 +863,7 @@ package object geotiff {
     tiffTile
   }
 
+  // This implementation is not used anymore
```
EmileSonneveld: Is it not used here? https://github.com/Open-EO/openeo-geopyspark-driver/blob/da0493f219db8b6c6817c60d7201eb8059bce060/openeogeotrellis/geopysparkdatacube.py#L2734

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/geotiff/package.scala
```diff
@@ -1007,7 +1032,11 @@ package object geotiff {
     if (fo.overviews.toUpperCase == "ALL" ||
       fo.overviews.toUpperCase == "AUTO" && (gridBounds.width > 1024 || gridBounds.height > 1024)
     ) {
-      geotiff = geotiff.withOverviews(getOverviewResampleMethod(fo), List(4, 8, 16), blockSize = fo.tileSize)
+      val baseOverview = geotiff.buildOverview(getOverviewResampleMethod(fo),2,blockSize = fo.tileSize)
```
dsamaey: extract getOverviewResampleMethod(fo) ?

dsamaey (approved):

PR #488, opened 2025-07-10
Keep metadata after resample

EmileSonneveld:
@bossie This did not change much in performance, but allows to keep metadata after a merge_cubes. Ok to merge?


jdries:
Was there a case that triggered the need for this? Or do we loose this metadata because we do resample_cube_spatial as part of merge_cubes??
In the current way of working, we simply set this metadata on the python side whenever we call a scala side process that needs it.

PR #492, opened 2025-07-23
fix: add support for empty NetCDF collections Ref: #476

### LINKED ISSUE #476: NetCDFCollection has no guard for empty stacItems list (allow_empty_cube)
https://github.com/Open-EO/openeo-geotrellis-extensions/blob/bd76938926dddb9cdefece57ae50739cef0abbcc/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/NetCDFCollection.scala#L39

Can throw:
```
java.util.NoSuchElementException: head of empty list
	at scala.collection.immutable.Nil$.head(List.scala:469)
	at scala.collection.immutable.Nil$.head(List.scala:466)
	at org.openeo.geotrellis.layers.NetCDFCollection$.loadCollection(NetCDFCollection.scala:39)
	at org.openeo.geotrellis.layers.NetCDFCollection$.datacube_seq(NetCDFCollection.scala:31)
	at org.openeo.geotrellis.layers.NetCDFCollection.datacube_seq(NetCDFCollection.scala)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at py4j.reflection.MethodInvoker.invoke(MethodInvoker.java:244)
	at py4j.reflection.ReflectionEngine.invoke(ReflectionEngine.java:374)
	at py4j.Gateway.invoke(Gateway.java:282)
	at py4j.commands.AbstractCommand.invokeMethod(AbstractCommand.java:132)
	at py4j.commands.CallCommand.execute(CallCommand.java:79)
	at py4j.ClientServerConnection.waitForCommands(ClientServerConnection.java:182)
	at py4j.ClientServerConnection.run(ClientServerConnection.java:106)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

While this should be possible when the `allow_empty_cube` feature flag is set.

[original pg](https://gist.github.com/JeroenVerstraelen/d5fe2a363eeffbb1dfecf7c9817346df)

dsamaey:


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/NetCDFCollection.scala
```diff
@@ -28,99 +27,130 @@ object NetCDFCollection {
...
-    val crs = items.flatMap(_.crs).distinct().collect()
-    if(crs.length != 1) {
-      throw new IllegalArgumentException("All items in a collection must have the same CRS")
-    }
+  private def loadEmptyCollection(polygons: ProjectedPolygons, from_date: String, dataCubeParameters: DataCubeParameters, sc: SparkContext) = {
+    val boundingBox = polygons.extent
 
-    val bboxWGS84: Extent = items.map(_.bbox).reduce((a, b)=>(a.combine(b)))
+    val from = ZonedDateTime.parse(from_date, ISO_OFFSET_DATE_TIME)
+    val to = ZonedDateTime.parse("2025-06-01T00:00Z", ISO_OFFSET_DATE_TIME)
```
EmileSonneveld: to_date can probably be passed as argument too

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/NetCDFCollection.scala
```diff
@@ -28,99 +27,130 @@ object NetCDFCollection {
...
 
+    val scheme = FloatingLayoutScheme(dataCubeParameters.tileSize)
+    val maxSpatialResolution = CellSize(10, 10)
```
EmileSonneveld: Would it make sense to get resolution from the stac metadata from stac?
For the data type, I am not sure if it is even available from the stac metadata

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/NetCDFCollection.scala
```diff
@@ -28,99 +27,130 @@ object NetCDFCollection {
...
-            if (!conventions.startsWith("CF-1")) {
-              throw new IllegalArgumentException(s"Only netCDF files with CF-1.x conventions are supported by this openEO backend, but found ${conventions}.")
-            }
-            if (extraDim != "{t}") {
-              throw new IllegalArgumentException("Only netCDF files with a time dimension named 't' are supported by this openEO backend.")
-            }
-            if( units != "days since 1990-01-01") {
-              throw new IllegalArgumentException("Only netCDF files with a time dimension in 'days since 1990-01-01' are supported by this openEO backend.")
+    if (stacItems.isEmpty) {
+      ContextRDD(sc.emptyRDD[(SpaceTimeKey, MultibandTile)], null)
```
EmileSonneveld: Would this be an ok RDD? If metadata is indeed optional, this might be an easier path compared to `loadEmptyCollection`

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/NetCDFCollection.scala
dsamaey: Done

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/NetCDFCollection.scala
jdries: makes sense, but not so easy to achieve: we want to move to dynamically getting resolution from the assets that are loaded, in this case there are no assets, so we would need to pass on some form of collection metadata info, from which resolution can be deducted.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/NetCDFCollection.scala
dsamaey: null is not an option

EmileSonneveld: Then it might be safer to throw an error here directly

PR #510, opened 2025-08-25
1234-save-to-geotiff-expose-compression-method-predictor-bigtiff

### LINKED ISSUE #1234: Save to Geotiff: expose compression method & predictor
Goal: For saving geotiffs, I would like to use ZSTD compression (instead of standard deflate), preferably along predictor=2.

Reference on both:
https://gdal.org/en/stable/drivers/raster/gtiff.html#:~:text=floating%20point%20values.-,COMPRESS,-%3D%5BJPEG%E2%80%8B/%E2%80%8BLZW%E2%80%8B/%E2%80%8BPACKBITS

Why:
ZSTD has been show to have a beneficial effect on both reading time and compression ratio vs. deflate.
See https://kokoalberti.com/articles/geotiff-compression-optimization-guide/

Proposal:
Expose these as as part of the save_results arguments:
- compression: string
- predictor: int
Possibly, it's preferred to limit compression arguments to zstd/deflate/lzw.

Priority: medium

dsamaey:


openeo-geotrellis/src/test/scala/org/openeo/geotrellis/layers/Sentinel2FileLayerProviderTest.scala
```diff
@@ -292,10 +292,10 @@ class Sentinel2FileLayerProviderTest extends RasterMatchers {
     builder.argumentEnd()
     builder.constantArgument("y", 6)
     builder.expressionEnd("gte", args)
-    mask.toSpatial(date).writeGeoTiff(f"tmp/Sentinel2FileLayerProvider_multiband_SCL_${parameters.hashCode()}.tif", bbox)
+    mask.toSpatial(date).writeGeoTiff(f"/tmp/Sentinel2FileLayerProvider_multiband_SCL_${parameters.hashCode()}.tif", bbox)
```
jdries: Should we construct a temp path in an OS independent way? (I would say either that, or else keep the relative path as-is for now.)

jdries (changes requested):

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/layers/Sentinel2FileLayerProviderTest.scala
dsamaey: half of this file already used "/tmp" and this failed locally, so I changed them all to "/tmp"

jdries: k, then it's consistent at least. I guess we'll want to add a temp path fixture in another PR.

jdries (approved):

EmileSonneveld:
@codex, review

PR #511, opened 2025-08-27
save_result: geotiff asset per band avoid copy

jdries:
Only write to temporary locations for extra task attempts. This way, we do not require an expensive copy of files in object storage.
Major performance improvement for LCFM.

@EmileSonneveld  can you review? Do you know if there was existing unit test coverage for this?
Also check: what happens if the destination path already exists when a file is copied? (This can now happen.)


EmileSonneveld:
Some errors came up from CI. I'll take a look and try to add a test that simulate the first attempt failing. (The big job that should reproduce it does not anymore)

EmileSonneveld:
Tested manually with letting the first attempt explicitly fail. By adding this snippet to the top of `writeTiff`
```scala
    if (TaskContext.get() != null && TaskContext.get().attemptNumber() == 0) {
      throw new RuntimeException("Test error on TaskContext.get().attemptNumber() == 0")
    }
```
Speculative execution settings don't seem to work in unit test environment.

Also added a test emulateting a failing executor writing a corrupt file that will be overwritten.

PR #540, opened 2025-09-22
minor: clarify load_stac in exception 

soxofaan:
Clarify in exception that  FileLayerProvider.loadRasterSourceRDD can also be triggered from `load_stac`

### DIFF
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
@@ -1603,7 +1603,7 @@ class FileLayerProvider private(openSearch: OpenSearchClient, openSearchCollecti
     // TODO: these geotiffs overlap a bit so for a bbox near the edge, not one but two or even four geotiffs are taken
     //  into account; it's more efficient to filter out the redundant ones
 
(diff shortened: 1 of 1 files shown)


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/layers/FileLayerProvider.scala
```diff
@@ -1603,7 +1603,7 @@ class FileLayerProvider private(openSearch: OpenSearchClient, openSearchCollecti
     // TODO: these geotiffs overlap a bit so for a bbox near the edge, not one but two or even four geotiffs are taken
     //  into account; it's more efficient to filter out the redundant ones
 
-    if (overlappingRasterSources.isEmpty) throw new IllegalArgumentException(s"""Could not find data for your load_collection request with catalog ID "$openSearchCollectionId". The catalog query had correlation ID "$correlationId" and returned ${overlappingFeatures.size} results.""")
+    if (overlappingRasterSources.isEmpty) throw new IllegalArgumentException(s"""Could not find data for your load_collection/load_stac/... request with catalog ID "$openSearchCollectionId". The catalog query had correlation ID "$correlationId" and returned ${overlappingFeatures.size} results.""")
```
bossie:
```suggestion
    if (overlappingRasterSources.isEmpty) throw new IllegalArgumentException(s"""Could not find data for your ${if (fromLoadStac) "load_stac" else "load_collection"} request with catalog ID "$openSearchCollectionId". The catalog query had correlation ID "$correlationId" and returned ${overlappingFeatures.size} results.""")
```

bossie (changes requested):
There's a `fromLoadStac` field that tells you the context in which the call was made so you can adapt the error accordingly.

To obtain this context, a cleaner way is probably to pass it explicitly instead of deriving it from the type of `OpenSearchClient` but yeah, tech debt.

PR #549, opened 2025-10-03
Spark 4.0.1

dsamaey:


bossie:
@dsamaey AFAICT this branch produces the `geotrellis-dependencies` jar  used in Terrascope's Python 3.11 images; ok to merge `develop` into this branch to keep it up-to-date?

jdries:
@bossie  normally this is not needed, there is a single dependencies jar that works for both spark versions.
(Specific Spark dependencies are installed directly in the image.)

bossie:
@jdries I'm referring to this line: https://github.com/Open-EO/openeo-geotrellis-kubernetes/blob/8ad31a688d91ee319ba60cadce50c3c942b0443c/docker/AlmaLinux/Jenkinsfile.almalinux9-spark4-py311-openeo#L12

jdries:
Looks like a leftover, can you try replacing that with the default jar, like:
https://artifactory.vgt.vito.be/artifactory/libs-snapshot-public/org/openeo/geotrellis-dependencies/2.6.1_2.13-SNAPSHOT/geotrellis-dependencies-2.6.1_2.13-SNAPSHOT.jar
(geopyspark unit tests should already have tested this approach, but of course not impossible that it breaks something.)

PR #575, opened 2025-11-05
feat: add process aggregate_spatial_window

### LINKED ISSUE #548: support aggregate_spatial_window
Process: https://processes.openeo.org/#aggregate_spatial_window

Use case: the 'multi-look' operation for coherence could use this. 

This operation is somewhat similar to resample_spatial, but with a user defined reducer.

In our own code, I'm not aware of process implementations that reduce spatial dimensions, but perhaps geotrellis does support them.

dsamaey:


clausmichele:
Are there news about this process? It would be important for the ClouDInSAR project.

@jdries @EmileSonneveld

EmileSonneveld:
@codex review

Issue #556, opened 2025-10-08
Error when zooming out in QGIS


EmileSonneveld:
[...]

PR #617, opened 2025-12-22
overview test tilesize

ElienVandermaesenVITO:
@dsamaey the check for the tilesize is not correct. I added a test which makes this clear
see #556


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/geotiff/package.scala
```diff
@@ -351,7 +351,7 @@ package object geotiff {
 
         decimationFactors.indices.toList
           .filter(i => {
-            val tileSizeIsBigEnough = math.min(tileLayout.tileCols, tileLayout.tileRows) >= decimationFactors(i)
```
ElienVandermaesenVITO, line 354: The overviews are already calculated and than you are going to delete it. It is better to filter them before creating them.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/geotiff/package.scala
ElienVandermaesenVITO: Do the filtering of the overviews here (see comment before)

openeo-geotrellis/src/test/scala/org/openeo/geotrellis/geotiff/WriteRDDToGeotiffTest.scala
```diff
@@ -614,6 +614,25 @@ class WriteRDDToGeotiffTest extends RasterMatchers {
     assertArrayEquals(croppedReference.toArray(), result2.band(0).toArrayTile().crop(2 * 256, 0, layoutCols * 256, layoutRows * 256).toArray())
   }
 
+
+  @Test
```
ElienVandermaesenVITO, line 618: This is a test that takes a lot longer than others (in jenkins 23s compared to 1-5s) . So I would not leave it like this. Try to improve it or ignore it.

PR #666, opened 2026-03-09
access: Use provider credentials

pvbouwel:
CFC credentials are usable across regions the region specific credentials were missing an underscore. By going for the provider credentials we allow cleanup of the misnamed ones.


bossie (approved):
LGTM

PR #677, opened 2026-03-24
#670 ClassCastException NetCDFWriter

### LINKED ISSUE #670: NetCDFRDDWriter ClassCastException due to unexpected celltype
I get the following error in the NetCDFRDDWriter when running load_collection for a certain extent in `TERRASCOPE_S2_TOC_V2`, most other extents work:
```
py4j.protocol.Py4JJavaError: An error occurred while calling z:org.openeo.geotrellis.netcdf.NetCDFRDDWriter.writeRasters.
: java.lang.ClassCastException: class [I cannot be cast to class [S ([I and [S are in module java.base of loader 'bootstrap')
	at ucar.nc2.jni.netcdf.Nc4Iosp.writeData(Nc4Iosp.java:3082)
	at ucar.nc2.jni.netcdf.Nc4Iosp.writeData(Nc4Iosp.java:2996)
	at ucar.nc2.NetcdfFileWriter.write(NetcdfFileWriter.java:1066)
	at ucar.nc2.NetcdfFileWriter.write(NetcdfFileWriter.java:1050)
	at org.openeo.geotrellis.netcdf.NetCDFRDDWriter$.writeTile(NetCDFRDDWriter.scala:358)
	at org.openeo.geotrellis.netcdf.NetCDFRDDWriter$.$anonfun$saveSingleNetCDFGeneric$5(NetCDFRDDWriter.scala:265)
	at scala.collection.immutable.Range.foreach$mVc$sp(Range.scala:192)
	at org.openeo.geotrellis.netcdf.NetCDFRDDWriter$.$anonfun$saveSingleNetCDFGeneric$3(NetCDFRDDWriter.scala:239)
	at org.openeo.geotrellis.netcdf.NetCDFRDDWriter$.$anonfun$saveSingleNetCDFGeneric$3$adapted(NetCDFRDDWriter.scala:202)
	at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:619)
	at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:617)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1306)
	at org.openeo.geotrellis.netcdf.NetCDFRDDWriter$.saveSingleNetCDFGeneric(NetCDFRDDWriter.scala:202)
	at org.openeo.geotrellis.netcdf.NetCDFRDDWriter$.saveSingleNetCDFGeneric(NetCDFRDDWriter.scala:150)
	at org.openeo.geotrellis.netcdf.NetCDFRDDWriter$.writeRasters(NetCDFRDDWriter.scala:81)
	at org.openeo.geotrellis.netcdf.NetCDFRDDWriter.writeRasters(NetCDFRDDWriter.scala)
```

I first thought this was related to the resolveTileOverlap code in the FileLayerProvider (see https://github.com/eu-cdse/openeo-cdse-infra/issues/956) but currently it happens regardless if this feature flag is on or not. It also happens for both the STAC and the Opensearch version of the collection.

## Job Ids
*  TERRASCOPE_S2_TOC_V2 on mep-prod: j-2603121439314bec8e42145f32d6f9a3
* TERRASCOPE_S2_TOC_V2 on mep-dev: j-2603121413064f66858c381e65bd3a71
* _TERRASCOPE_S2_TOC_V2_OPENSEARCH on mep-dev: j-2603121322264a82b025baf72ef8d244 

## Reproduction
You can reproduce the issue with the following code. The batch job is expected to fail for both collections.
https://gist.github.com/JeroenVerstraelen/d04f0349e5cb2f001a8007f2e0a4365f

## Priority
This issue is currently on production for a very popular layer so this has high priority.

ElienVandermaesenVITO:
#670

### DIFF
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/netcdf/NetCDFRDDWriter.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/netcdf/NetCDFRDDWriter.scala
@@ -199,7 +199,7 @@ object NetCDFRDDWriter {
       }
     val bandStatistics = collection.mutable.Map[String,(Double,Double,Double,Double,Int,Int)]() // min, max, sum, sum of the power, valid data count, total data count
     var netcdfFile: NetcdfFileWriter = null
-    val cellType = preProcessedRdd.metadata.cellType
+    val cellType = cachedRDD.map(x => x._2.cellType).fold(BitCellType)((x,y)=>x.union(y))
     for(tuple <- cachedRDD.toLocalIterator){
 
       val timeDimIndex =
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/netcdf/NetCDFRDDWriter.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/netcdf/NetCDFRDDWriter.scala
@@ -262,7 +262,6 @@ object NetCDFRDDWriter {
 
             var tile = multibandTile.band(bandIndex)
             if (tile.cellType != cellType) {
-              logger.info(s"before conversion: band $variable: original cellType of tile: ${tile.cellType}, needs to be converted to $cellType")
               tile = tile.convert(cellType)
             }
 
@@ -365,7 +364,6 @@ object NetCDFRDDWriter {
         case t: DoubleArrayTile => ucar.ma2.Array.factory(DataType.DOUBLE, shape, t.array)
       }
 
-    logger.info(s"variable: $variable, origin: ${origin.mkString("Array(", ", ", ")")}, type of bandArray: ${bandArray.getDataType}, and shape: ${shape.mkString("Array(", ", ", ")")}")
     netcdfFile.write(variable, origin, bandArray)
   }
 
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/netcdf/NetCDFRDDWriter.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/netcdf/NetCDFRDDWriter.scala
@@ -253,7 +253,9 @@ object NetCDFRDDWriter {
 
             var tile = multibandTile.band(bandIndex)
             if (tile.cellType != cellType) {
+              logger.info(s"before conversion: band $variable: original cellType of tile: ${tile.cellType}, needs to be converted to $cellType")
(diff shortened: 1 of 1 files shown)


openeo-geotrellis/src/main/scala/org/openeo/geotrellis/netcdf/NetCDFRDDWriter.scala
```diff
@@ -199,9 +199,18 @@ object NetCDFRDDWriter {
       }
     val bandStatistics = collection.mutable.Map[String,(Double,Double,Double,Double,Int,Int)]() // min, max, sum, sum of the power, valid data count, total data count
     var netcdfFile: NetcdfFileWriter = null
+    val cellType = cachedRDD.map(x => x._2.cellType).fold(BitCellType)((x,y)=> {
```
bossie: Is this functionally different from: https://github.com/Open-EO/openeo-geotrellis-extensions/blob/a894f2651637656e54b207b94768be451f1238fb/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/OpenEOProcessScriptBuilder.scala#L85-L103?

Could you move this to a named function?

bossie (changes requested):
Could you add a simple unit test that mimics the minimal process graph from #670?

AFAICT there's no more difference for the `SENTINEL3_SLSTR_L2_LST` process graph so 👍

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/netcdf/NetCDFRDDWriter.scala
ElienVandermaesenVITO: For the possible` CellTypes` they will give the same result. The `cellTypeUnion` considers a case where there is an unsigned and a floating point with the same bits, which does not exists. So it is just a different order of checking things

ElienVandermaesenVITO: I'm hoping that geotrellis corrects its union function, so that we can use that one. (see the added comment)

bossie: Ok, but in the meanwhile, can we just keep one implementation then, whichever you think is best?

PR #755, opened 2026-05-26
A new Healpix module

jdries:
This PR adds a new type of datacube, for the HealPix DGGS representation.

It currently has 2 experimental representations: either one row for each cell, or a packed variant where multiple cells belonging to the same parent are stored as an array. The purpose is to enable performance prototyping.

This datacube is based on Spark SQL dataframes rather than regular the RDD's used by Geotrellis. This should give access to advanced features such as adaptive query execution, which may reduce the need for custom partitioner tuning in our code.

The new representation also comes with introspection capabilities for the available processes. The goal is to allow new openEO processes to be implemented in Scala, without requiring Python code to be added in all cases.

Additionally, a Sentinel-3 reader and Zarr writer are added. The ucar netcdf library needed an update because it can now read from object storage directly.

**First impression of Healpix/DGGS compatibility with openEO**

The Healpix 'nested' variant is essentially a quadtree organization, with cell id's following a Z-order. This matches very well with the spatial indexes used by e.g. Geotrellis. The result is that the external Healpix model will very closely follow the internal representation in our backend, making it easier for developers and users to understand the behavior.

Processes on individual cells or groups of cells that do not depend on a spatial relationship (reduce_dimention, apply_dimension,...) should be rather compatible, and many may continue to work 'as-is'. This PR also adds an adapter to make this possible.

Processes like apply_neighborhood and apply_kernel will probably not be applicable, an we'll need to come up with new DGGS capable variants. An apply_neighborhood variant where the neighborhood is specified by a shared parent cell id would for instance be a very 'natural' operation on a DGGS.

Collection-wise, for medium to low resolution datasets (Sentinel-3, Sentinel-5p), it will become much easier to implement correct large-area processing, without being implicitly affected by projection issues. 

**This code is AI generated, with some initial review done already, plus testing via unit and intergration tests. The purpose is to set up an MVP, not production code. We want to quickly experiment to assess feasibility of a full DGGS backend.**


openeo-geotrellis-healpix/src/main/scala/org/openeo/geotrellishealpix/HealpixDataGenerator.scala
```diff
@@ -0,0 +1,190 @@
...
+      val latRad = math.Pi / 2.0 - p.theta
+      Row.fromSeq(Seq[Any](cid, ts, math.sin(latRad)))
+    }
+    val df = spark.createDataFrame(spark.sparkContext.parallelize(rows), schema)
+    ScalarHealpixDatacube(nside, bandDefs, df)
+  }
+
+  // ---- Java-friendly overloads (using java.util.List and ISO 8601 strings) ----
+
+  /**   * Java-friendly version of randomScalar using java.util.List and ISO 8601 timestamp strings.   *   * @param spark active SparkSession   * @param nside HEALPix NSIDE (power of two)   * @param timestamps list of ISO 8601 timestamp strings (e.g., "2024-05-27T10:30:00Z")   * @param bands list of band/variable names   * @param seed random seed (default: 42)   * @return ScalarHealpixDatacube with random values   */
```
jdries: @copilot  these javadoc strings are all on a single line, while they should be multiline
same applies to other javadoc in this file

PR #756, opened 2026-05-26
S3proxy support

pvbouwel:
Allow configuration of clients against the S3proxy by providing a bucket configuration file.

Since these codes path fallback to the previous client this should have limited impact but it allows:
- Providing access paths without code changes in the future
- toggling s3proxy usage at job level

### DIFF
--- a/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/creo/CreoS3Utils.scala
+++ b/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/creo/CreoS3Utils.scala
@@ -1,11 +1,12 @@
 package org.openeo.geotrellis.creo
 
+import com.fasterxml.jackson.databind.ObjectMapper
 import geotrellis.store.s3.AmazonS3URI
 import org.apache.commons.io.FileUtils
 import org.apache.commons.io.filefilter.TrueFileFilter
 import org.openeo.geotrelliss3.S3Utils
 import org.slf4j.LoggerFactory
-import software.amazon.awssdk.auth.credentials.{AwsBasicCredentials, StaticCredentialsProvider}
+import software.amazon.awssdk.auth.credentials.{AnonymousCredentialsProvider, AwsBasicCredentials, StaticCredentialsProvider}
 import software.amazon.awssdk.awscore.retry.conditions.RetryOnErrorCodeCondition
 import software.amazon.awssdk.core.client.config.ClientOverrideConfiguration
 import software.amazon.awssdk.core.retry.RetryPolicy
@@ -15,6 +16,8 @@ import software.amazon.awssdk.core.sync.RequestBody
 import software.amazon.awssdk.regions.Region
 import software.amazon.awssdk.services.s3.model._
 import software.amazon.awssdk.services.s3.{S3AsyncClient, S3Client, S3Configuration}
+import software.amazon.awssdk.services.sts.StsClient
+import software.amazon.awssdk.services.sts.auth.StsWebIdentityTokenFileCredentialsProvider
 import software.amazon.awssdk.transfer.s3.S3TransferManager
 import software.amazon.awssdk.transfer.s3.model.UploadFileRequest
 
@@ -27,6 +30,7 @@ import scala.util.control.Breaks.{break, breakable}
 
 object CreoS3Utils {
   private val logger = LoggerFactory.getLogger(getClass)
+  private val objectMapper = new ObjectMapper()
 
   private val cloudFerroRegion: Region = Region.of("RegionOne")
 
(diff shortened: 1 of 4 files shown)


EmileSonneveld:
@copilot review without making changes

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/creo/CreoS3Utils.scala
```diff
@@ -47,6 +54,79 @@ object CreoS3Utils {
...
+  }
+
+  private def buildProxyS3Client(bucketName: String): S3Client = {
+    val tokenFile = Path.of(sys.env.getOrElse("OPENEO_WEB_IDENTITY_TOKEN_FILE", "/opt/job_config/token"))
+    if (!Files.isRegularFile(tokenFile) || !Files.isReadable(tokenFile)) {
+      logger.warn(s"Cannot build proxy S3 client for bucket $bucketName: web identity token file is not readable: $tokenFile")
+      return null
+    }
+
+    val bucketConfigFile = Path.of(sys.env.getOrElse("OPENEO_BUCKET_CONFIG_FILE", "/opt/job_config/bucket_config.json"))
```
EmileSonneveld, line 72: Is there an example `bucket_config.json` would look like?

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/creo/CreoS3Utils.scala
```diff
@@ -104,9 +184,69 @@ object CreoS3Utils {
...
+        null
+    }
+  }
+
+  private def proxyRoleSessionName(bucketName: String): String = {
+    val podName = sys.env.getOrElse("SPARK_EXECUTOR_POD_NAME", "unknown-sparkappid")
+    val sparkAppId = podName.split("-").slice(0, 2).mkString("-")
+    val executorId = sys.env.getOrElse("SPARK_EXECUTOR_ID", "00")
+
+    s"$sparkAppId-$executorId-$bucketName".take(64)
```
EmileSonneveld, line 244: Does the session name need to be unique? We have some buckets that are 64 characters long, and postfix to the name might get cut off. 
https://docs.aws.amazon.com/sdk-for-swift/latest/api/awssts/documentation/awssts/assumeroleinput/rolesessionname/

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/creo/CreoS3Utils.scala
```diff
@@ -47,6 +54,79 @@ object CreoS3Utils {
...
+        null
+    }
+  }
+
+  def getS3Client(uri: AmazonS3URI): S3Client = {
+    val proxy = getProxyS3Client(uri.getBucket)
+    if (proxy != null) proxy else getCreoS3Client()
+  }
+
+  //Prefer using getS3Client with an S3 URI
```
EmileSonneveld, line 129: Might be cleaner:
`@deprecated("Prefer using getS3Client with an S3 URI")`
But then the `getCreoS3Client` calls in `rangeReader` would need to be updated too

dsamaey (approved):

EmileSonneveld (commented):
Is there a ticket where the mr can link to?

EmileSonneveld (approved):
LGTM. If there would be something wrong, the logic would fallback to what it was

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/creo/CreoS3Utils.scala
pvbouwel: Generally sparkappid + executor id is what makes it moslty unique. Bucket name is something that could be interesting to see. Theoretically if one executor indeed interacts with multiple buckets with a similar prefix then it could be cut-off but that would just make it a bit more cumbersome. In our S3 proxy we also have AKID logged per request so that can still be used to uniquely identify which session is used.
Depending on the use case session names are unique but it is not required nor enforced by STS.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/creo/CreoS3Utils.scala
pvbouwel: Indeed, that is not in scope of this PR but would rather be part of a cleanup phase once logic is in place to always have bucket config managed structurally by the Python driver at which point the calls in rangeReader make sense to be replaced. The logic is to now allow specifying bucket JSON. To eventually have that bucket JSON as being mandatory such that region determination is always decoupled from geotrellis-extensions.

openeo-geotrellis/src/main/scala/org/openeo/geotrellis/creo/CreoS3Utils.scala
pvbouwel: That will be documented at the python side but in the end it will look something like:
```json
{
  "bucket1": {
    "region": "region-a",
    "role_arn": "arn:openeows:iam:::role/workspace1"
  },
  "bucket2": {
    "region": "region-b",
    "role_arn": "arn:openeows:iam:::role/workspace2"
  },
  ...
}
```

Which is a processed form of https://github.com/Open-EO/openeo-python-driver/blob/master/openeo_driver/config/config.py#L96

PR #778, opened 2026-06-10
Geocoding for SAR cube support

jdries:
Experimental process, but useful to merge in as this is an important capability.


jdries:
@EmileSonneveld  is it corret that you disabled sentinelhub support on this branch?
I would like to merge it in, but that might be hard without sentinelhub. Can you restore it, or provide more detail on what was the issue?

EmileSonneveld:
Trying something out with copilot. https://github.com/Open-EO/openeo-geotrellis-extensions/issues/779

PR #789, opened 2026-06-15
A generated module to compute sigma0 backscatter directly

jdries:
Open-EO/openeo-geopyspark-driver#921

This new module provides a pure scala implementation for sigma0 backscatter.

While it adds a lot of code, it would remove the dependency on orfeo toolbox, and allows to be more memory efficient because it can use the jvm heap memory.

It also allows native support for reading all files from S3, which would simplify the backend setup. (e.g. no /vsis3 mount hack).

It also opens the door to experiment with gamm0 RTC approximations rather than waiting for one of the dependencies to implement it.

TODO:

- [ ] changelog
- [ ] DEM handling: can we provide the DEM as a cube, or is there a better approach?
- [ ] geoid file
- [ ] performance evaluation on backend
- [ ] support old products with border noise
- [ ] check if products with corrupt data are supported (see hack we needed in orfeo)


jdries:
I set up a collection on dev, that loads from GRD stac catalog.
The new raster source provider will pick this up automatically using default backscatter settings.

```
    c = openeo.connect("openeo.dev.warsaw.openeo.dataspace.copernicus.eu").authenticate_oidc()
    s2_cube = c.load_collection(
        "SENTINEL1_GRD_SIGMA0",
        temporal_extent = ("2024-05-07", "2024-05-10"),
        spatial_extent = {
            "east": -68.590184,
            "south": -22.412385,
            "west": -68.939554,  
            "north": -22.318986, 
            "crs": "EPSG:4326"
        },
        bands = ['vv'],
    )
    job_options = {
        "stac-version": "1.1",
        "derived-from-document-experimental": True,
        "openeo-jar-path": "https://artifactory.vgt.vito.be/artifactory/libs-release-public/org/openeo/geotrellis-extensions/PR-789/geotrellis-extensions-PR-789.jar"
    }
    s2_cube.execute_batch("chili3_cdse.nc", title= "sigma0 test", log_level="debug",job_options=job_options)
```

jdries:
Status update: performance is slower than OTB, likely because VV and VH computation is entirely repeated due to how FileLayerProvider currently treats multi-band raster sources.
Otherwise, the implementation seems to work quite well already. I'm considering to merge this PR, and then make progress on the FileLayerProvider issue.

PR #790, opened 2026-06-15
framework and end to end test for collection loading

jdries:
This PR proposes an end-to-end approach for testing of loading complex collections:

- Test data is generated on disk, no large binaries.
- Generated collections are retained across tests for performance
- Generated collections can easily be inspected.
- Test output is also written to disk: allows easy inspection
- The test can remain valid under large refactorings, because we test at interface level
- The test has no external dependencies


jdries:
@dsamaey I'll be picking up this one, as I depend on it to merge other work.
Do let me know if there are comments on the current code, otherwise I'll probably merge when I have a green build.

jdries:
Current failure: 
org.opentest4j.AssertionFailedError: B11 must be larger on the right (XGradient): left sum=2491666.591796875, right sum=0.0 ==> expected: <true> but was: <false>

This one is caused by an actual 'known' issue, the test is good at triggering specific cases. Need to see how to deal with it.

